PR #2, opened 2016-12-22
Initial release 2N Intercom devices

pfischi:
This is the first release for 2N Intercom devices.

http://www.keil-telecom.de/produkte/ip-sprechanlagen/


intercom_2n/utils.py
```diff
@@ -0,0 +1,33 @@
...
+                    break
+                except IOError:
+                    pass
+    except socket.gaierror:
+        return get_lan_ip_fallback()
+    return ip
+
+def get_interface_ip(ifname):
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    return socket.inet_ntoa(fcntl.ioctl(s.fileno(), 0x8915, struct.pack('256s', ifname[:15].encode('utf-8')))[20:24])
```
ohinckel: Is here a socket cleanup required to free some resources? Just to avoid a possible memory leak...

intercom_2n/utils.py
```diff
@@ -0,0 +1,33 @@
+import os
+import socket
+import struct
+import fcntl
+
+def get_lan_ip():
+    try:
+        ip = socket.gethostbyname(socket.gethostname())
+        if ip.startswith("127.") and os.name != "nt":
+            interfaces = ["eth0", "eth1", "eth2", "wlan0", "wlan1", "wifi0", "ath0", "ath1", "ppp0"]
```
ohinckel: What if users will not have such interfaces and use different names or having multiple LAN interfaces. Could also be a configuration option for the plugin, right?

intercom_2n/utils.py
```diff
@@ -0,0 +1,33 @@
...
+def get_interface_ip(ifname):
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    return socket.inet_ntoa(fcntl.ioctl(s.fileno(), 0x8915, struct.pack('256s', ifname[:15].encode('utf-8')))[20:24])
+
+def get_lan_ip_fallback():
+    try:
+        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
+        s.connect(('<broadcast>', 0))
+        return s.getsockname()[0]
```
pfischi: Same as above.

intercom_2n/utils.py
pfischi: You're right. I'll add an option to the plugin config.

intercom_2n/utils.py
pfischi: Don't think so. the soket.inet_ntoa function convert an IP address from 32-bit packed binary format to string format. The socket object should be destroyed after leaving the function. There should be no other reference left.

pfischi: I'll rewrite this code fragment. Should be easy to fix.

pfischi: The problem is gone :)
I can just delete the utils.py. I've used it because i had first implemented a http-listener. This is now removed because i using the more stable pulli mode.

pfischi:
utils.py is removed.

ohinckel:
Great! Thanks for your contribution!

PR #9, opened 2017-01-08
changes uzsu to Smartplugin (single instance only)

bmxp:


uzsu/__init__.py
```diff
@@ -46,52 +46,98 @@
...
+        :return:        If the plugin needs to be informed of an items change you should return a call back function
+                        like the function update_item down below. An example when this is needed is the knx plugin
+                        where parse_item returns the update_item function when the attribute knx_send is found.
+                        This means that when the items value is about to be updated, the call back function is called
+                        with the item, caller, source and dest as arguments and in case of the knx plugin the value
+                        can be sent to the knx with a knx write function within the knx plugin.
+
+        """
+	
+        if 'uzsu_item' in item.conf:
```
ohinckel: What about checking this using `has_iattr` of `SmartPlugin`?

uzsu/__init__.py
```diff
@@ -46,52 +46,98 @@
...
+    
+    ALLOW_MULTIINSTANCE = False
+    
+    PLUGIN_VERSION = "1.3.0"
+
 
     _items = {}         # item buffer for all uzsu enabled items
 
-    def __init__(self, smarthome, path=None):
+    def __init__(self, smarthome, path=None, *args, **kwargs):
```
ohinckel: Is there any reason why this `args` or `kwargs` arguments were added? AFAIS they are not used somewhere.

uzsu/__init__.py
```diff
@@ -46,52 +46,98 @@
...
+        """
+        Initalizes the plugin. The parameters describe for this method are pulled from the entry in plugin.conf.
+
+        :param sh:  The instance of the smarthome object, save it for later references
```
ohinckel: Additionally the documentation for `path` parameter can be added, too. Just to make it complete.

uzsu/__init__.py
bmxp: Doesn't hurt and it might get options one day :-)

uzsu/__init__.py
bmxp: Good idea.

uzsu/__init__.py
bmxp: Since it is single instance we don't need it right now but to make it more complete, we have to overwork the whole plugin. I just wanted to implement the most needed things...

PR #10, opened 2017-01-15
Add helios plugin

ohinckel:
As requested by @Tom-Bom-badil I created a branch for the helios plugin he created. This PR will integrate the helios plugin keeping the original history.


Tom-Bom-badil:
Thanks a lot @ohinckel, commits can be deleted if still possible, so we have a clean Version '1'.

PR #23, opened 2017-02-19
XBMC: added shutdown command

smaugs:
added shutdown command. one can now use xbmc_send = shutdown in items.conf

### DIFF
--- a/xbmc/__init__.py
+++ b/xbmc/__init__.py
@@ -54,7 +54,7 @@ class xbmc(lib.connection.Client):
 
     _notification_time = 10000
     _listen_keys = ['volume', 'mute', 'title', 'media', 'state']
-    _send_keys = {'volume': 'Application.SetVolume', 'mute': 'Application.SetMute'}
+    _send_keys = {'volume': 'Application.SetVolume', 'mute': 'Application.SetMute', 'shutdown': 'System.Shutdown'}
 
     def __init__(self, smarthome, item):
         if 'xbmc_port' in item.conf:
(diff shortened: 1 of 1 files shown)


msinn:
Could you document the additioal feature in the README.md?

onkelandy:
Any chance to update the plugin furthermore? Right now it's very rudimentary.. or maybe I just don't get it how to use it? You can only send volume and mute command, right?

ohinckel:
Adding the new command to the documentation would be really nice. If added we could merge it.

@onkelandy, yes only these commands were supported. Which commands do you need? I think supporting the whole system is really much work. So we should first - if somebody want - focus on most used or important commands.

onkelandy:
Play, pause, stop would be the minimum for me. As a matter of fact I've implemented stop, but I had problems with implementing the play command somehow (can't remember why). 

Showing the currently playing title would also be nice - as well as the power commands (esp. restart)
Then an implementation in the VISU would start to make sense ;)

Thanks a lot.

smaugs:
Sorry I was pretty busy the last weeks - I'll document the shutdown command this weekend. Maybe I'll have a look at stop/play - see if I have the same problem(s) as onkelandy :)

ohinckel:
Updating the documentation would be enough in the first step to merge this PR. Please update the documentation and create a new PR for the additional commands.

ohinckel:
@smaugs already updated the documentation? After this we can merge and close this PR. Thanks for your support!

psilo909:
also waiting for a documentation update. .yaml examples in the README would be nice

also a migration to SmartPlugin and the recommended use of the logger as attribute would be nice

onkelandy:
@psilo909 This one could be killed because it's replaced by:
https://github.com/smarthomeNG/plugins/pull/95

psilo909:
@onkelandy what do you mean with killed? close the PR?

onkelandy:
The kodi plugin replaces this XBMC plugin.. so actually the whole XBMC plugin could/should be removed.

ohinckel:
Merged the current state of this now into `develop`. Further development should go to new `kodi` plugin - see #95.

PR #31, opened 2017-04-02
UTE learn mode added: can be called in interactive mode via "start_UT…

beckerth:
…E_learnmode(id_offset)", if UTE teach in request is received plugin will answer automatically and teach in the device with the given offset id.

### DIFF
--- a/enocean/__init__.py
+++ b/enocean/__init__.py
@@ -205,7 +205,15 @@ class EnOcean(SmartPlugin):
             dBm = -optional[5]
             SecurityLevel = optional[6]
             self.logger.debug("enocean: radio message with additional info: subtelnum = {} / dest_id = {:08X} / signal = {}dBm / SecurityLevel = {}".format(subtelnum, dest_id, dBm, SecurityLevel))
-
+            if (choice == 0xD4) and (self.UTE_listen == True):
+                #self.logger.info("UTE FOUND!! RORG {:02x}".format(choice))
+                self.logger.info("call send_UTE_response")
+                self._send_UTE_response(data, optional)
+                #self.logger.debug("send response")
+                #logger.debug("time in milliseconds")
+                #logger.debug(datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S.%f')[:-3])
+                #self._send_radio_packet(5, choice, [0x91, 0xFF, 0x33, 0x00, 0x09, 0x01, 0xD2],[0x03, 0x01, 0x93, 0xAF, 0x59, 0xFF, 0x0] )#UTE response package for psc234 - 0193AF59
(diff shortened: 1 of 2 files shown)


enocean/__init__.py
```diff
@@ -205,7 +205,15 @@ def _process_packet_type_radio(self, data, optional):
...
             self.logger.debug("enocean: radio message with additional info: subtelnum = {} / dest_id = {:08X} / signal = {}dBm / SecurityLevel = {}".format(subtelnum, dest_id, dBm, SecurityLevel))
-
+            if (choice == 0xD4) and (self.UTE_listen == True):
+                #self.logger.info("UTE FOUND!! RORG {:02x}".format(choice))
+                self.logger.info("call send_UTE_response")
+                self._send_UTE_response(data, optional)
+                #self.logger.debug("send response")
+                #logger.debug("time in milliseconds")
+                #logger.debug(datetime.utcnow().strftime('%Y-%m-%d %H:%M:%S.%f')[:-3])
+                #self._send_radio_packet(5, choice, [0x91, 0xFF, 0x33, 0x00, 0x09, 0x01, 0xD2],[0x03, 0x01, 0x93, 0xAF, 0x59, 0xFF, 0x0] )#UTE response package for psc234 - 0193AF59
```
aschwith: What is this commented line for? Looks like you were trying something, which is fine but please remove for the pul request unless it is important.

enocean/__init__.py
```diff
@@ -205,7 +205,15 @@ def _process_packet_type_radio(self, data, optional):
...
+                #self._send_radio_packet(8, choice, [0x91, 0x02, 0x46, 0x00, 0x12, 0x01, 0xD2],[0x03, 0x01, 0xA6, 0x20, 0x2B, 0xFF, 0x0] )#UTE response package for Nodon actor - 01A6202B
```
aschwith: Same here, what is this commented line for? Looks like you were trying something, which is fine but please remove for the pul request unless it is important.

enocean/__init__.py
```diff
@@ -669,7 +696,67 @@ def send_learn_switch(self, id_offset=0):
             return
         self.logger.info("enocean: sending learn telegram for switch command")
         self._send_radio_packet(id_offset, 0xA5, [0x01, 0x00, 0x00, 0x00])
-
+
```
aschwith: The content of the functions below are fine but please combine those into one single on/off function, i.e. see send_switch function prototype. Additionally, please not not use device specific names, like nodon. That way we would have thousands of different functions for Eltako, Nodon, Gira etc. Use the enocean way instead and name them by EEP. Thanks!

beckerth: I'm working on it. This is already removed in my commits.

aschwith:
Thanks for incorporating the changes. Accept and close.

PR #34, opened 2017-04-11
Squeezebox plugin: added automatic request of playlist name

onkelandy:
playlist name gets polled when new tracks are loaded in playlist.


ohinckel:
Merging this, since it only contains **new** commands which will not affect any other working code.

PR #39, opened 2017-05-01
package xmltodict added

pfischi:
python package xmltodict added


pfischi:
Please do a re-scan for all requirements to update the requirements.txt in the SmarthomeNG requirement folder.

PR #40, opened 2017-05-05
Add logic trigger to operationlog

ohinckel:
This patch will add support to configure logic using `operationlog` plugin to generate a log message each time the logic was triggered.

The feature is not new, it was previously implemented in the `memlog` plugin. Since the operationlog is a little bit more useable (can log to file, can cache the log, ...) I would also suggest to completely remove the `memlog` plugin later (not part of this PR).

To configure such a message for a logic trigger you can use the following settings in the `logic.conf`:
* olog = \<name\>
* olog_txt = "...."
* olog_level = \<level\>

PS: I added myself to the copyright notice since I think parts and ideas are from the original `memlog` plugin. Hope it's OK. If not, let me know.

### DIFF
--- a/operationlog/README.md
+++ b/operationlog/README.md
@@ -125,6 +125,40 @@ The code is replaced by the return value of the \<python code> for the logtext. 
         olog_level = info
 ```
 
+## Logics
+Configure a logic to be logged as follows:
+
+```
+[some_logic]
+    filename = script.py
+    olog = mylogname1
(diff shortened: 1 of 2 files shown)


msinn:
I like this addition. 

Now only one thing is missing to throw out my own logging plugin: I need to log an item-change to more than one logfile at a time.

ohinckel:
Could be solved by making it a multi instance plugin.

PR #45, opened 2017-05-11
fix stamp

kr2:


ohinckel:
Thanks for fixing this! But the `timestamp()` method is only available in Python 3.3+. Currently we still support version 3.2.

kr2:
Is there some place where I can read up on that and why you do it? On my first skim through the wiki I didn’t see anything about supporting python 3.2. 
German is fine too.

msinn:
There is a quit long running discussion on backward compatibility.

Take a look at [issue 34](https://github.com/smarthomeNG/smarthome/issues/34) (Support für Python 3.2 und Python 3.4)

It hasn't be finally decided, but Python backward compatibility will tend to 'actual version and two versions back) as a general line.

ohinckel:
Just one more note: the README describes the `time` field as "the string representation of the time" which is not meant to be a timestamp. Maybe introducing a new field would be better? Beside of `time` we could also support `timestamp`.

ohinckel:
@kr2 can you please have a look at this issue? I would merge this because we do not support Python 3.2 anymore starting with 1.3 release, but I still prefer to use a new field instead of changing an exisiting one. User can decide if date/time string or timestamp should be logged.

PR #48, opened 2017-05-29
NUT - new plugin

4d4mu:
Hi,
Recently I created this new plugin to connect with NUT - Network UPS Tools daemon. So far works fine with my server, now it's time to share :) I tried to make it SmartPlugin and multiinstance (to work with more than one ups device).
Can anyone test it? :)

4d4mu

### DIFF
--- a/nut/README.md
+++ b/nut/README.md
@@ -1,0 +1,30 @@
+# NUT - Network UPS Tool plugin
+
+Requirements
+============
+none
+
+
+Configuration
+=============
+
+## plugin.conf
+
+<pre>
+ [nut]
+
+    class_name = NUT
+    class_path = plugins.nut
+    host = localhost
+    port = port
+    cycle = cycle
+</pre>
+
+## items.conf
+<pre>
+ [ups_item]
+      
+    type = str
+    nut_var = variable
+</pre>
+
--- a/nut/README.md
+++ b/nut/README.md
@@ -1,8 +1,9 @@
 # NUT - Network UPS Tool plugin
+This plugin is connecting to NUT daemon and can be used to read ups variables. You can list available variables using command `upsc your_ups_name`.
 
 Requirements
 ============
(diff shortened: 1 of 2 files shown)


psilo909:
unfortunately cant test ist.. where you could modify is the plugin version. you are writing 1.3.0, the last digit in fact is the plugin version, so 1.3.0.1 would be plugin version 0.1.. version "0" is perhaps a bit misleading

psilo909:
thx... in my opinion i could just merge it.. but it would be good if someone could test it before (i cant :-/)

4d4mu:
Thanks. Actually, nut allows to create dummy devices from text files, so anyone with nut-compatible OS and some time and goodwill can test it :) I've tested multi-instance mode with one real device and one dummy. Or if I'm the only user right now, I guess we can merge it

psilo909:
well give it some more days, but being the only user also accounts for some of my plugins ;-) at least i think so.. so even if there are bugs, these can be discussed when you are not the only user anymore

psilo909:
ah and sorry that i 4got about: a support thread in the knx user forum reference from the readme as link and an entry in https://github.com/smarthomeNG/smarthome/wiki/Liste-der-Plugins (after the plugin has been merged) would be nice. in case u are not using the forum, i dont mind

4d4mu:
Currently I'm not using this forum. Is it all German or English posts are allowed?
BTW, i found similar plugin - apcups. NUT is compatible with many APC devices, so maybe someone can test it... @cstrassburg?

ohinckel:
If there are no other users I think we can merge this. I guess it will not break anything. A support thread linked in the readme file would be nice!

psilo909:
as he is not using the forum, i dont think we can demand that thread.. merging is ok for me.

4d4mu:
thanks

PR #51, opened 2017-06-03
alexa alias-names & turn-on/off values (knx-scene-support) & new lock-states

hotzen:


hotzen:
now into develop and also extended

PR #52, opened 2017-06-06
hue: fix group command 'on'

r4vn:
Using the `on` command with groups results in an error message complaining about a missing lamp id. This change fixes this issue and swaps the item store variable of group command `on`. 

I'm aware that the target of this pull request is the `master` branch. This is due to changes which are missing on the `develop` branch.

### DIFF
--- a/hue/__init__.py
+++ b/hue/__init__.py
@@ -466,10 +466,10 @@ class HUE():
                 # lampe ist an (status in sh). dann können alle befehle gesendet werden
                 if hueSendGroup == 'on':
                     # wenn der status in sh true ist, aber mit dem befehl on, dann muss die lampe auf der hue seite erst eingeschaltet werden
-                    if hueIndex + '.bri' in self._sendLampItems:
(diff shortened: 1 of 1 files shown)


ohinckel:
This patch looks simple enough to be applied to `develop` branch too. Did you test applying it to develop branch?

Btw. the difference between `develop` and `master` branch is only one commit https://github.com/smarthomeNG/plugins/commit/7976c3152304ced684ac0fdcf3f857efa7327a92#diff-fe57e2b3b242b2df5b53e27f466c264f

In general I don't like commiting to the `master` branch since everything should be in `develop` first - even bug fixes. But others here maybe handle this differently.

r4vn:
I also applied it to `develop` but not actually test it. However I could open a pull request for `develop` if you want to.

ohinckel:
Would be nice if you can provide a new PR based on `develop` branch. Usually everything should go first into the `develop` branch.

PR #54, opened 2017-06-18
GPIO plugin update

onkelandy:
Changed GPIO detection to GPIO.event_detect which makes the plugin much more streamlined.


ohinckel:
Nice update for the gpio plugin using callbacks to update the items. I would like to test it, but it seems that this PR contains other changes too.

Do you think you can create another PR having only changes for the gpio plugin?

onkelandy:
github macht mich wahnsinnig.. Ich versuch's nochmals frisch. Sorry.

PR #55, opened 2017-06-18
Updated plugin GPIO

onkelandy:
Change to GPIO.event_detect to streamline plugin

### DIFF
--- a/gpio/__init__.py
+++ b/gpio/__init__.py
@@ -23,25 +23,16 @@ import logging
 from lib.model.smartplugin import SmartPlugin
 import RPi.GPIO as GPIO
 import threading
-import time
-from time import sleep
 
 class Raspi_GPIO(SmartPlugin):
-    PLUGIN_VERSION = "1.0.0"
+    PLUGIN_VERSION = "1.0.1"
     ALLOW_MULTIINSTANCE = False
 
-    def _wait(self,time_lapse):
-        time_start = time.time()
-        time_end = (time_start + time_lapse)
-     
-        while time_end > time.time():
-            pass
-
-    def __init__(self, sh, cycle=0, mode="board"):
+    def __init__(self, sh, mode="board"):
         self.logger = logging.getLogger(__name__)
         self._sh = sh
         self._items = []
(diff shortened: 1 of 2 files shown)


gpio/__init__.py
```diff
@@ -79,23 +74,30 @@ def parse_item(self, item):
             GPIO.setup(in_pin, GPIO.IN, pull_up_down=GPIO.PUD_UP)
             self.logger.debug("GPIO: INPUT {0} assigned to pin \'{1}\'".format(item, in_pin))
             self._items.append(item)
+            self._itemsdict[in_pin] = item
             return self.update_item      
         if self.has_iattr(item.conf, 'gpio_out'):
             out_pin = int(self.get_iattr_value(item.conf, 'gpio_out'))
+            GPIO.setup(out_pin, GPIO.IN, pull_up_down=GPIO.PUD_UP)
```
ohinckel: Is this just a copy&paste error or do I need to use `setup()`  with `GPIO.IN` and later with `GPIO.OUT`?

gpio/__init__.py
```diff
@@ -79,23 +74,30 @@ def parse_item(self, item):
...
+            self._itemsdict[out_pin] = item
+            value = GPIO.input(out_pin)
+            item(value, 'GPIO', 'gpio_init')
+            GPIO.add_event_detect(out_pin, GPIO.BOTH, callback=self.get_sensors)
```
ohinckel: I'm not sure, but does this event detection works with output pins too?

gpio/__init__.py
onkelandy: The GPIO.IN is necessary to detect value changes in the output pin. Something like a "reply" whether it worked or not

gpio/__init__.py
onkelandy: As far as I see it, it works for out pin but I am not totally sure. At least it is possible to query the current value of an output pin with GPIO.input

ohinckel:
Just tested the `gpio_in` with the event decetion implementation: it's working and load is reduced. Thanks for looking into this @onkelandy!

PR #57, opened 2017-06-24
Update README.md

henfri:
Hallo, 

ich habe den Parameter cycle zur Readme hinzugefügt und den Bereich der items für die Vorhersage korrigiert.

Gruß,
Hendrik

### DIFF
--- a/wunderground/README.md
+++ b/wunderground/README.md
@@ -32,6 +32,8 @@ You can configure multiple instances of the wunderground plugin to collect data 
 	location = Germany/Hamburg
 	item_subtree = mein_wetter
 	instance = wetter_ham
+	wug_datatype = positive
+	#cycle = 600
 ```
 
 or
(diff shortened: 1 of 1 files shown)


ohinckel:
Really want to merge this into `master`?

msinn:
This PR contains a couple of errors:

- wug_datatype for plugin.yaml / plugin.conf: This makes no sense
- for forcast values the forcast days are missing
- forcast values only defined for the current day and not for the days to come

henfri:
> wug_datatype for plugin.yaml / plugin.conf: This makes no sense

Why?

>for forcast values the forcast days are missing
Yes. No change to the current status.
I just corrected/clarified the rain probability vs. the rain amount.

msinn:
Short answer: Because it is not implemented.
Long answer: wug_datatype is implemented as an attribute for individual items. So, it can be used in items.ymal / items.conf. It is not a plugin parameter. So it is not to be used in plugin.yaml / plugin.conf.

What should it do in the plugin.yaml?

PR #61, opened 2017-07-01
Kostal piko plugin with json support

datenschuft:
added JSON-communication to the kostal plugin.
since ui-firmware version 6 json-format is supported


kostal/README.md
```diff
@@ -1,14 +1,21 @@
...
+VERSION = "1.3.1"
+
 # Requirements
 
-This plugin is designed to retrieve data from a [KOSTAL](http://www.kostal-solar-electric.com/) inverter module (e.g. PICO inverters).
+This plugin is designed to retrieve data from a [KOSTAL](http://www.kostal-solar-electric.com/) inverter module (e.g. PIKO inverters)
+Since UI-version 6 json-format is supported
+
+Tested with Kostal Piko 3.0 (single phase)  UI-Version 06.20
+dc-line 2, ac line 2 and ac line 3 values unchecked/untested
```
ohinckel: I would kike to see the list of possible values in the configuration section. And we should mentioned that the "old" PIKO converters are still supported with the html format, new ones with the JSON format.

kostal/README.md
```diff
@@ -1,14 +1,21 @@
...
 
 ## Supported Hardware
 
-Is currently working with the following KOSTA inverter modules:
+Is currently working with the following KOSTAL inverter modules:
 
-  * KOSTAL PIKO 7.0
+  * KOSTAL PIKO 3.0
+  (should work with all KOSTAL PIKO inverters)
```
ohinckel: Maybe we can change this to: Supported converters:
* PIKO 7.0 (old ones)
* PIKO 3.0 (new ones, next generation)

Additionally we can link to the home pages:
old ones http://kostal-solar-electric.com/de-DE/Produkte_Service/PIKO-Wechselrichter_bewaehrte_Generation,
new ones http://kostal-solar-electric.com/de-DE/Produkte_Service/PIKO-Wechselrichter_neue_Generation

kostal/README.md
```diff
@@ -17,43 +24,35 @@ Is currently working with the following KOSTA inverter modules:
...
-
-  * power_current: The current power of the solar installation
-  * power_total: The total amount of generated energy
-  * power_day: The amount of generated energy for the current day
-  * status: The textual status of the module (off, feeding, ...)
-  * string1_volt ... string3_volt: The current voltage of string 1, 2, 3
-  * string1_ampere ... string3_ampere: The current ampere of string 1, 2, 3
-  * l1_volt ... l3_volt: The current voltage of L 1, 2, 3
-  * l1_watt ... l3_watt: The current watt of L 1, 2, 3
-
```
ohinckel: As mentioned before I would like to see a list of possible values for both the html and json format. If possible we could also use the same keys (e.g. `l1_volt` instead of `dc1_v`). In this case we only have one list of supported values and the datastructure doesn't matter - maybe easier for one having the items configured for an old converter and want to switch to a new converter by just change the `datastructure` setting.

kostal/README.md
```diff
@@ -63,35 +62,97 @@ daily power. Additionally it shows the volts and watts for the phases.
...
+        kostal = yield_day_wh
+    [[yield_tot_kwh]]
+        name = Gesamtleistung total
+        type = num
+        kostal = yield_tot_kwh
+    [[operationtime_h]]
+        name = Betriebsstunden
+        type = num
+        kostal = operationtime_h
 </pre>
```
ohinckel: If we do not want to use the same keys/names for `kostal` setting we should provide an example for both types (html and json).

kostal/__init__.py
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
 # vim: set encoding=utf-8 tabstop=4 softtabstop=4 shiftwidth=4 expandtab
-#
-# Copyright 2013 KNX-User-Forum e.V.            http://knx-user-forum.de/
-#
-#  This file is part of SmartHomeNG.    https://github.com/smarthomeNG//
+#########################################################################
+#  Copyright 2017 Wenger Florian                       <wenger@unifox.at>
```
ohinckel: Can you pleas add an additional copyright line:
`#  Copyright 2012-     Oliver Hinckel                  github@ollisnet.de`

kostal/__init__.py
```diff
@@ -18,88 +18,182 @@
...
-            for item_cfg in self._items:
-                if item_cfg[1] in self._values:
-                    item_cfg[0](self._values[item_cfg[1]], 'Kostal')
-        except Exception as e:
-            logger.error(
-                'could not retrieve data from {0}: {1}'.format(self.ip, e))
-            return
-
+        if self.datastucture == "html":
+            #HTML-OLD-Coding
```
ohinckel: I would move the code for the old and new datastructure handling into separate methods. Additionally I would assign this method to the `datastructure` attribute instead of a simple string (see init method) and simply invoke it `self.datastructure()` (this will the invoke the assigned method from init method).

This would also shrink the method _refresh().

kostal/__init__.py
```diff
@@ -18,88 +18,182 @@
...
         cycletime = time.time() - start
-        logger.debug("cycle takes {0} seconds".format(cycletime))
+        self.logger.debug("cycle takes {0} seconds".format(cycletime))
+
+
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.DEBUG, format='%(relativeCreated)6d %(threadName)s %(message)s')
+    # todo
+    # change PluginClassName appropriately
+    PluginClassName(None).run()
```
ohinckel: Remove this main code block or update the plugin class name.

kostal/__init__.py
```diff
@@ -18,88 +18,182 @@
...
         pass
-
     def update_item(self, item, caller=None, source=None, dest=None):
-        if caller != 'Kostal':
-            pass
-
+        if item():
+            if self.has_iattr(item.conf, 'kostal'):
+                self.logger.debug("update_item ws called with item '{}' from caller '{}', source '{}' and dest '{}'".format(item, caller, source, dest))
+                pass
```
ohinckel: Can we remove the code block in update_item()? I think we do not need to implement anything there.

kostal/README.md
```diff
@@ -1,14 +1,21 @@
 # KOSTAL
 
+VERSION = "1.3.1"
+
 # Requirements
```
datenschuft: Pleas create a support thread... I'll add the link to README.md.

kostal/__init__.py
datenschuft: I don't know why update_item is needed. But it is defined in the sample_plugin/__init__.py 
https://github.com/smarthomeNG/smarthome/blob/master/dev/sample_plugin/__init__.py
I thought it is needed by smarthomeng.

ohinckel: You're right. The example is just a template for the plugin, but I don't think that all plugin should implement and keep this method like shown in this example. It should help to understand what all the methods do.

I would implement the method like this:
```
    def update_item(self, item, caller=None, source=None, dest=None):
        """
        Write items values
        :param item: item to be updated towards the plugin
        :param caller: if given it represents the callers name
        :param source: if given it represents the source
        :param dest: if given it represents the dest
        """
        pass
```

kostal/__init__.py
```diff
@@ -128,72 +132,67 @@ def parse_item(self, item):
...
-                                value = "unknown"
-                        if kostal_key in self._items:
-                            self._items[kostal_key](value)
-                            self.logger.debug("items[" + str(kostal_key) +"] = " +str(value))
-            except Exception as e:
-                self.logger.error(
-                    'could not retrieve data from {0}: {1}'.format(self.ip, e))
-                return
+        # run the working methods
+        self.datastructure()
```
ohinckel: General exception handling could be removed from the different datastructure metehods and moved to this place to have it only one time. But it's only sugar - so not really required.

ohinckel:
Cool, thanks to take care about all my suggestion!

kostal/__init__.py
datenschuft: i think it is good, if each method handle its own exceptions

kostal/README.md
ohinckel: Just created a new thread for this plugin. URL is:
https://knx-user-forum.de/forum/supportforen/smarthome-py/1109697-kostal-plugin-piko-wechselrichter

kostal/README.md
```diff
@@ -1,97 +1,201 @@
...
-Is currently working with the following KOSTA inverter modules:
+Is currently working with the following KOSTAL inverter modules:
+
+  * KOSTAL PIKO 3.0 UI-Version 06.20 (datastructure=json)
+  * KOSTAL PIKO 5.5 UI-Version 05.xx (datastructure=html)
+
+  (should work with all KOSTAL PIKO inverters)
+  <add more successfull testet Kostal Inverters with UI-Version and used datastructure>
+
+knx-user-forum.de
```
ohinckel: Was this just a typo or why is the knxuf domain inserted here?

kostal/__init__.py
```diff
@@ -18,88 +19,190 @@
...
-        'string3_ampere': 102,
-        'l1_volt': 48,
-        'l2_volt': 72,
-        'l3_volt': 96,
-        'l1_watt': 57,
-        'l2_watt': 81,
-        'l3_watt': 105
+        'actot_w': 9,
+        'yield_tot_kwh': 12,
+        'yield_day_wh': 21,
```
ohinckel: The value for `yield_day_wh` in the HTML version is not `Wh`, it is `kWh` too:
```
Energie
Gesamtenergie 33386 kWh
Tagesenergie  26.58 kWh
```
I think we need a scale to specify the conversion (or unit-conversion methods) for generating always the same units for different devices.

kostal/README.md
```diff
@@ -1,97 +1,201 @@
...
+        type = num
+        kostal = ac3_v
+    [[ac3_a]]
+        name = Phase 3 Strom
+        type = num
+        kostal = ac3_a
+    [[ac3_w]]
+        name = Phase 3 Leistung
+        type = num
+        kostal = ac3_w
```
ohinckel: I don't see such a value for the HTML version, but it could easily calculated by the other two values (`ac3_a` and `ac3_v`). Do you think it make sense to support as most as possible values for both types of converters?

ohinckel:
Thanks for this greate update!

Can you have a look at the conflicts? Just rebase your work on an updated develop branch (it seems that the README.md file was [modified](https://github.com/smarthomeNG/plugins/commit/e583e7054d0eab199ff2607279c3d08b0cfd6828#diff-a68f915590f1e7898f782bac2d2de00e) to fix the header styles - required to generate the documentation correctly).

PR #65, opened 2017-08-13
Initial version of Pushover Plugin

ThomasCr:
Hi, this is my initial version of Pushover Plugin.

### DIFF
--- a/pushover/README.md
+++ b/pushover/README.md
@@ -65,7 +65,7 @@ Send a message to your device.
 * __userKey__: defines and/or override the globally defined user-key of the receiving user
 * __apiKey__: defines and/or override the globally defined api-key of the sending application
 
-  All params can set to None (only message not), so they will not be set or in case of device, userKey and apiKey the global vars will be used.
+All params can set to None (only message not), so they will not be set or in case of device, userKey and apiKey the global vars will be used.
   
 #### Example
 ```python
--- a/pushover/README.md
+++ b/pushover/README.md
@@ -1,12 +1,14 @@
 # Pushover  
-  
-## Requirements  
 
+Pushover is one out of many push services, which is compatible with Android, IOS and Windows Clients. This plugin make it possibe to push messages to your clients from smarthomeNG.
+  
+## Requirements
+  
 ### Python libraries  
 * requests - [install instructions](http://docs.python-requests.org/en/latest/user/install/#install "http://docs.python-requests.org/en/latest/user/install/#install")
   
 ### Other  
-* Pushover API-KEY - get it from [__here__](https://pushover.net/apps/ "https://pushover.net/apps/") for free  
+* Pushover API-KEY - get it from [__here__](https://pushover.net/apps/ "https://pushover.net/apps/") for free, after registration.
   
 ---
 ## Changelog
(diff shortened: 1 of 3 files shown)


psilo909:
please align the plugin version to https://github.com/smarthomeNG/smarthome/wiki/Versionsnummern

e.g. 1.3.0.1 (version 0.1 of plugin, usable with smarthomeng 1.3)

we will build functionality based on that, so it needs to be correct

psilo909:
it would also be good to shortly describe what the plugin does (before requirements) and to set a support thread in the knx user forum, which is then linked from readme

psilo909:
oh and in the yaml example in the readme you use the "=" in several lines

pushover/README.md
```diff
@@ -0,0 +1,85 @@
+# Pushover  
+
+Pushover is one out of many push services, which is compatible with Android, IOS and Windows Clients. This plugin make it possible to push messages to your clients from smarthomeNG.
```
ohinckel: SmartHomeNG should be with capital S,H and NG.

ThomasCr:
hope it is okay now ;)

pushover/README.md
```diff
@@ -0,0 +1,85 @@
...
+  
+## Requirements
+  
+### Python libraries  
+* requests - [install instructions](http://docs.python-requests.org/en/latest/user/install/#install "http://docs.python-requests.org/en/latest/user/install/#install")
```
ohinckel: Do we need to mention the requirements here, when we have them listed in the requirements file?

pushover/README.md
ThomasCr: fixed

pushover/README.md
ThomasCr: dont know.. possible it will help users which install the requirement manually?

psilo909: mentioning is ok, but install instructions are perhaps a bit too much.. people need to use pip for smarthomeng in any way

ohinckel: I would list it if we list them in other plugins too. On the other hand one can check the requirements file to know which packages are required.

pushover/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+        if apiKey:
+            data['token'] = apiKey
+        else:
+            data['token'] = self._apiKey
+
+        if device:
+            data['device'] = device
+        elif self._device:
+            data['device'] = self._device
+
```
ohinckel: Does sending the message works when apikey/userkey/device is `None` (because it was not specified in `__init__()` method and not in call of this method)?

ThomasCr: no, apikey and userkey need to be defined. but error message get automatically dropped in request, because params are needed.

ThomasCr: also when it is not essential, I have added a check for it ;)

ThomasCr:
It is now okay?

pushover/README.md
msinn: I think someone wanting to use a plugin should find the requirements (hard- and software) in one place (the README.md). The requirements section doesn't need to specify the exact version(s). It can refer to the requirements file.

I think a lot of users don't now Python very well (of not at all) and have no clue about the requirements file. If the requirement of a module is documented in the README with a reference to the requirements file (regarding to the version), all user would be helped.

PR #67, opened 2017-08-15
Develop

SergeoLacruz:
Added new function to filter events by their caller

### DIFF
--- a/simulation/README.md
+++ b/simulation/README.md
@@ -28,9 +28,15 @@ simulation:
     class_name: Simulation
     class_path: plugins.simulation
     data_file: /usr/smarthome/var/db/simulation.txt
+    callers = knx | visu
 ```
 
 `data_file`: This is the file where all recorded events are stored.
+
+`callers`: is a list of event sources for recording of events. When an item is changed, the change is done
+by someone, e.g. knx for changes from the bus. The canner name is identical with the plugin name that changes the item.
+Only item changes with a caller in the list are recorded to the simulation file. In the example above e.g. uzsu is
+ignored.
 
 ### items.conf (deprecated) / items.yaml
 
--- a/simulation/README.md
+++ b/simulation/README.md
@@ -21,6 +21,7 @@ This plugins has no requirements.
    class_name = Simulation
    class_path = plugins.simulation
    data_file = /usr/smarthome/var/db/simulation.txt
+   callers = knx | visu
 ```
 
 ```yaml
--- a/simulation/README.md
+++ b/simulation/README.md
@@ -34,7 +34,8 @@ simulation:
 `data_file`: This is the file where all recorded events are stored.
 
 `callers`: is a list of event sources for recording of events. When an item is changed, the change is done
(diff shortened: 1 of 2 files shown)


simulation/README.md
```diff
@@ -32,6 +33,11 @@ simulation:
 
 `data_file`: This is the file where all recorded events are stored.
 
+`callers`: is a list of event sources for recording of events. When an item is changed, the change is done
+by someone, e.g. knx for changes from the bus. The canner name is identical with the plugin name that changes the item.
```
ohinckel: Typo in `canner`.

ohinckel: I think the `caller` is not the plugin name. It is something the plugin developer decide. The KNX plugin is using `KNX`, but registering a plugin multiple times you are not able to ecide if it was from the first or second plugin instance.

Usually the plugin updates the item by just calling `item(value, caller, ...)`. The caller parameter can be anything. Just an idea: we could introduce a method in plugins superclass which updates an item and always uses the plugin instance name as caller value.

SergeoLacruz: Hi,
You are right. The programmer sets the caller name. Even I did it in the plugin :-)
Just did not remember.
I corrected the description. 

Michael

psilo909:
@ohinckel, @SergeoLacruz i had to do some corrections

callers needed a default in the init method, otherwise existing configurations will break.
also the readme was not updated concerning the YAML example.

ohinckel:
An empty list as callers default value would be better, which does not restrict something at all - which was the old behaviour.

psilo909:
@ohinckel wont this fail?

if (item.conf['sim'] == 'track') and (self.state()==2) and (caller in self._callers)

SergeoLacruz:
A default behaviour that records everything would be preferred but en empty list 
will record nothing. 
In addition the list content is case sensitive. So the documantation is still not OK.
I will work on the default behaviour and correct the doc

psilo909:
just rename it to caller_filter and implement that an empty list means all are allowed? ;)

SergeoLacruz:
Unfortunately not so easy. When the list is empty nothing will be recorded. 
So it is a calle_enabler. Not a good idea. 
I think this ie easy to fix. 
But actually fork does not work. I deleted the branch after the merge. When I now push Fork 
in smarthomeNG/plugins nothing happens.

ohinckel:
Btw. I just [implemented](https://github.com/smarthomeNG/plugins/commit/ce6230cc313b784f91787f35558b02cd674e8c20): the `callers` option is now optional. If not configured everything is recorded as before. Otherwise, if given, it will only record matching plugins.

I think this is OK now. Maybe the name `callers` can be adjusted to clarify that this is a "restriction filter". But the current version is OK for me.

SergeoLacruz:
The code is OK with your change.
Anyhow we should mention in the doc that the list is case sensitive and correct the
examples to knx and visu to KNX and Visu

psilo909:
I just updated the readme. Should be ok now. Thanks for the fixes! I also documented the plugin changes on our Release 1.4 page.

PR #70, opened 2017-09-11
Update __init__.py

DrGEH:
Incorporated the simple electronic switch D2_01_07 with an additional parameter pulsewidth to switch off the actuator after a certain time.

### DIFF
--- a/enocean/__init__.py
+++ b/enocean/__init__.py
@@ -441,7 +441,7 @@ class EnOcean(SmartPlugin):
 
     def _send_UTE_response(self, data, optional):
         choice = data[0]
-        #payload = data[1:-5]
+        payload = data[1:-5]
         #sender_id = int.from_bytes(data[-5:-1], byteorder='big', signed=False)
         #status = data[-1]
         #repeater_cnt = status & 0x0F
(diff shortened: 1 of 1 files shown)


enocean/__init__.py
```diff
@@ -441,7 +441,7 @@ def start_UTE_learnmode(self, id_offset=0):
 
     def _send_UTE_response(self, data, optional):
         choice = data[0]
-        #payload = data[1:-5]
+        payload = data[1:-5]
```
ohinckel: This looks like an important fix, since the `payload` variable is use later. Without this fix the plugin method failed due to undefined variable, right? Was it ever woriking?

enocean/__init__.py
```diff
@@ -526,6 +526,22 @@ def update_item(self, item, caller=None, source=None, dest=None):
...
+                            self.logger.info('enocean:  NO enocean_rx_id found')
+                        if 'pulsewidth' in item.conf:
+                            pulsew = float(item.conf['pulsewidth'])
+                            self.logger.info('enocean:  pulsewidth found')
+                        else:
+                            pulsew=0
+                            self.logger.info('enocean:  NO pulsewidth found')
+                        self.logger.info('enocean: item is D2_01_07_01 type')
+                        self.send_switch_D2(id_offset, rx_id, pulsew, item())
+                        self.logger.debug('enocean: sent switch command for D2 VLD')
```
ohinckel: Do you really want to log this stuff above using `INFO` level? I would prefer to use `DEBUG` instead.

ohinckel:
The documentation in the `README.md` for thisnattribute is missing. Can you add it?

I saw that the other attributes are also not documented. Would be nice if someone reading the documentation knows which attribute can be set and what it does.

Attributes missing (hopefully complete listing) :
- enocean_id
- enocean_rorg
- enocean_value
- enocean_rocker_action
- enocean_rocker_sequence
- enocean_rx_id
- enocean_rx_key
- enocean_rx_eep
- enocean_tx_eep
- enocean_tx_id_offset
- ref_level
- pulsewidth
- enocean_rtime

enocean/__init__.py
DrGEH: I doubt that it was working before. At least in my case, it wasn't.  After removing the hash signed at least it did work..

enocean/__init__.py
DrGEH: You're right, DEBUG is better. Could you please change it.

ohinckel: I can't change it. Just change it in your local copy, then commit to branch and push it again. The PR will be updated then.

DrGEH: I just changed the INFO into DEBUG. Furthermore, I renamed pulsewidth to enocean_pulsewidth so that we are more consistent with the names of the attributes.

DrGEH:
I updated README.md and __init__.py. Hope that you can see both commits for __init__.py.

PR #71, opened 2017-09-11
Update eep_parser.py

DrGEH:
Added D2-01-07


ohinckel:
Can you provide a more descriptive description? I would like to megre it but I don't know what I should write into the release notes.

DrGEH:
Support added for protocol D2-01-07 which is a simple electronic switch like Peha  d 451 fu-ebi pf o.t. nr. 00364476. The status of the actuator (on/off) is returned in the variable STAT.

PR #72, opened 2017-09-15
make use_checksum actually work again

hotzen:
to support smartmeters with bogus checksums like itron ace3000.

most changes caused by atom.io cleaning up spaces and stuff. actual change is only line 409

### DIFF
--- a/dlms/__init__.py
+++ b/dlms/__init__.py
@@ -39,7 +39,7 @@ if __name__ == '__main__':
         pass
 else:
     from lib.model.smartplugin import SmartPlugin
-        
+
 import time
 import serial
 import re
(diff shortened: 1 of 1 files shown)


bmxp:
Hm. I see. To be honest, I would to prefer havein an opt out instead of using ``use_checksum`` as opt-in.
IMHO something like ``suppress_checksum_generation`` expresses the normal assumption that a checksum should always be used and that the user intentionally doe not want to use the checksum for data check.

hotzen:
since this is an parameter introduced ages ago, I would not rename it now. I merely changed its functionality from fully broken to working-again.
despite this, I fully agree with you. suppress would be much more sensible.

ohinckel:
The arguments were removed in https://github.com/hotzen/smarthome-plugins/commit/9e0328816e1b838c3a2e5a22b6a58bdeab2386ff and added again https://github.com/smarthomeNG/plugins/commit/9e0328816e1b838c3a2e5a22b6a58bdeab2386ff#diff-d038b38ef8ed40d90927e9810a467fd0 to be backward compatible again and avoid upgrading problems.

We should decide to completely remove them or completely support it again.

bmxp:
Ok, since the default is to use the checksum we will leave it as the PR suggests it. So one needs to switch it intentionally off to skip the checksum part.
@ohinckel please go ahead and merge it.

ohinckel:
Thanks for your opinion @bmxp! I'll merge it...

PR #75, opened 2017-09-18
visu_smartvisu: fix category menu on mobile devices

4d4mu:
By default, category site contained `category_nav` in `sidebar block` and bunch of broken links in German in `content block`. Since there's no sidebar in smaller resolutions (like mobile phones), there's no other access to category item than manually editing address. This commit changes category behavior to act similar to rooms:
 -  `category_nav` in `content block` when selected category from menu; clock & weather in `sidebar block`
-  `category_nav` in `sidebar block` when selected specific category from  `category_nav`

### DIFF
--- a/visu_smartvisu/tplNG/category.html
+++ b/visu_smartvisu/tplNG/category.html
@@ -8,33 +8,8 @@
 */
 
 
-{% extends "base.html" %}
-
-{% block sidebar %}
-	{% include 'category_nav.html' %}
-{% endblock %}
+{% extends "index.html" %}
 
 {% block content %}
-
-	Category Content
-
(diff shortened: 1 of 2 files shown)


4d4mu:
@ohinckel, could You test it?

ohinckel:
What should I test here? Did you mean #234?

4d4mu:
I ment this pr. Or maybe you can point somebody to test & pull? #234 also would be great, but AFAIU it's blocked by decision what to do with plugins that need path.

msinn:
It has a small impact on non-mobile (or tablet) devices, but that's ok: When clicking on category. instead of displaying the category menu on the left and the first category details oh the right, now it behaves like the main menu (weather on the left, category menu on the right).

PR #80, opened 2017-10-15
MPD: added support for loading and playing local playlists

E3EAT:
added support for loading and playing local playlists which means seding the following commands to MPD to play a playlist:

clear
load filename
play


ohinckel:
This PR is against the `master` branch. Please create a new one against the `develop` branch.

PR #81, opened 2017-10-16
Version 1.3.1

onkelandy:
PEP8 optimized code
updated plugin.yaml and README
updated reset function for Denon receivers when no correct response received
updated automatic optimization of send command list
added debug verbose levels
full zone 2 and 3 control possible
added a lot of new commands to Denon text file

### DIFF
--- a/avdevice/__init__.py
+++ b/avdevice/__init__.py
@@ -36,23 +36,28 @@ from itertools import groupby
 
 import serial
 import socket
+from .AVDeviceInit import Init
+
+import sys
+sys.path.append('/usr/local/lib/python3.5/dist-packages/pysrc')
+import pydevd
 
 
 class AVDevice(SmartPlugin):
     ALLOW_MULTIINSTANCE = True
-    PLUGIN_VERSION = "1.3.0"
+    PLUGIN_VERSION = "1.3.1"
 
     def __init__(self,
                  smarthome,
                  model,
                  manufacturer='',
-                 ignoreresponse='',
+                 ignore_response='',
                  forcebuffer='',
                  inputignoredisplay='',
                  dependson=None,
                  rs232=None,
                  tcp=None,
-                 errorresponse='E02,E04,E06',
+                 error_response='E02,E04,E06',
                  resetonerror=False,
                  depend0_power0=False,
                  depend0_volume0=False,
@@ -60,9 +65,9 @@ class AVDevice(SmartPlugin):
                  resendwait=1.0,
                  reconnectretries=13,
                  secondstokeep=50,
-                 reconnectcycle=10,
                  responsebuffer='-5',
-                 autoreconnect=False
+                 autoreconnect=False,
+                 verboselevel=0
                  ):
         self.logger = logging.getLogger(__name__)
         self._sh = smarthome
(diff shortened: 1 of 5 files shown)


psilo909:
@onkelandy This branch has conflicts that must be resolved

avdevice/__init__.py
```diff
@@ -36,33 +36,38 @@
 
 import serial
 import socket
+from .AVDeviceInit import Init
```
ohinckel: I suggest to move the "local" import to the end of the standard module imports.

avdevice/__init__.py
```diff
@@ -36,33 +36,38 @@
...
+
+import sys
+sys.path.append('/usr/local/lib/python3.5/dist-packages/pysrc')
+import pydevd
```
ohinckel: I guess `pydevd` module was only used while developing the plugin, right? The static path added to `sys.path` too, I guess.

avdevice/__init__.py
```diff
@@ -469,136 +122,238 @@ def _wait(self, time_lapse):
...
+                    self.logger.debug("Resetting {}: Searching for suiting command for sendingcommand response: {}.".format(self._name, response))
+            except Exception as e:
+                response = self._send_commands[0].split(",")[2].split("|")
+                sendcommand = self._send_commands[0]
+                if self._verboselevel >= 2:
+                    self.logger.warning("Resetting {}: Cannot find Sendingcommand. Using first Command in queue: {}. Message: {}".format(self._name, response, e))
+            sorted_response_commands = sorted(self._response_commands, key=len, reverse=True)
+            for key in sorted_response_commands:
+                if self._verboselevel >= 2:
+                    self.logger.debug("Resetting {}: Trying to reset Item. Comparing: {} with {}".format(self._name, key, response))
```
ohinckel: Does a check on `verboselevel > XY` makes sense? I would avoid such things and instead use the common logger configuration to configure debug logging.

avdevice/plugin.yaml
```diff
@@ -0,0 +1,164 @@
...
+        description:
+            de: "Ein Wert zwischen 0 und 3, um das Verboselevel des Debugloggers zu definieren. Je höher, desto mehr Nachrichten werden geloggt."
+            en: "Value between 0 and 3 to define the verbose level of the debug logger. The higher the value, the more debug messages."
+  
+item_attributes:
+    avdevice_zone[0-4]@[instance]:
+        type: depending on command
+        description:
+            de: 'Obiges Attribut muss mit dem Kommando ergänzt werden, das in der Text Datei hinterlegt ist, zB power oder volume'
+            en: 'This attribute has to be followed by one of the commands declared in the device text file like power or volume'
```
ohinckel: I think we should not include the `@[instance]` parameter to the list of item attributes. It's redundant since the plugin is a multi instance plugin, and I don't know if we want to use the list of items (later) to validate parameters. @msinn can tell more about this feature.

Further we need to decide if we also want to support such names using placeholders/ranges like `name[0..9]` or `name[a-z]`. We could use the attribute's name directly or a regular expression (which support ranges) which will then be used for attributes matching the name expression.

What do you think @msinn?

onkelandy:
shit, sorry, was too eager to commit ;) have to delete pydev of course first.. I'll redo the stuff and make a clean pull request. Maybe I can also fix other issues. So ignore this PR or use it for further discussion ;)

Concerning verbose level: how can I do that using logging.yaml? I want to have different levels of debug, not only info/debug.. 

type can actually be any string that has to be listed in a separate file. What should I put in there?

What about the other stuff in plugin.yaml? Guess there is also some type, etc. to fix?

msinn:
> Concerning verbose level: how can I do that using logging.yaml? I want to have different levels of debug, not only info/debug..

Take a look at [Logging Best Practices](https://github.com/smarthomeNG/smarthome/wiki/Logging----Best-Practices). It shows how you can log different modules, plugins, etc. with different loglevels to a seperate log file, leaving your main logfile neat and clean.

onkelandy:
still don't get it how to define the specific log level in the plugin itself.
See https://github.com/smarthomeNG/plugins/pull/82#issuecomment-337217135

P.S.: the link to best practices is broken here: https://www.smarthomeng.de/dev/logging.html

msinn:
Your idea from https://github.com/smarthomeNG/plugins/pull/82#issuecomment-337217135 wouldn't work. The Python logging module does not work that way.

you can use:

self.logger.<level>("...)

But the levels are predefined by the Python logging mechanism. They are DEBUG, INFO, WARNING, ERROR, CRITICAL. You cannot define your own ones.

msinn:
There is another logging function: self.logger.log(<level>, "...") where level is an integer and the logging module defines constants with the levels (e.g.: logging.WARNING, ...)

PR #82, opened 2017-10-17
AVDevice: version 1.3.1

onkelandy:
PEP8 optimized code
updated plugin.yaml and README
updated reset function for Denon receivers when no correct response received
updated automatic optimization of send command list
added debug verbose levels
full zone 2 and 3 control possible
added a lot of new commands to Denon text file

### DIFF
--- a/avdevice/plugin.yaml
+++ b/avdevice/plugin.yaml
@@ -3,23 +3,162 @@ plugin:
     # Global plugin attributes
     type: interface                 # plugin type (gateway, interface, protocol, system, web)
     description:
-        de: 'Steuerung diversen AV Geräten über RS232 Schnittstelle, **seit SmartHomeNG v1.3**'
-        en: ''
+        de: 'Steuerung diversen AV Geräten über TCP/IP und RS232 Schnittstelle, **seit SmartHomeNG v1.3**'
+        en: 'Controlling AV devices via TCP/IP and RS232'
     maintainer: onkelandy
-#    tester:                        # Who tests this plugin?
-#    keywords: iot xyz
+    tester: Foxi352                     # Who tests this plugin?
+    keywords: av denon pioneer epson amp receiver projector rs232 telnet tcpip
 #    documentation: https://github.com/smarthomeNG/smarthome/wiki/CLI-Plugin        # url of documentation (wiki) page
-#    support: https://knx-user-forum.de/forum/supportforen/smarthome-py
+    support: https://knx-user-forum.de/forum/supportforen/smarthome-py/1097870-neues-plugin-av-device-f%C3%BCr-yamaha-pioneer-denon-etc
 
-    version: 1.3.0                 # Plugin version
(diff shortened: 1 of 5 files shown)


onkelandy:
Newly created request, able to merge...

Concerning verbose level: how can I do that using logging.yaml? I want to have different levels of debug, not only info/debug..

bmxp:
The logging level like logging.DEBUG or logging.CRITICAL are essentially just integers. You can define and use own levels: like DEBUGVERBOSE = 9 or VERYDEBUGVERBOSE = 1

onkelandy:
Didn't really get it.. How do I use that?
self.logger.verbose1("Blabla")
and in logging.yaml I assign level: verbose1?

The link here to the best practice https://www.smarthomeng.de/dev/logging.html#id6 actually doesn't lead anywhere :(

bmxp:
Link to best practices can be found [here](https://github.com/smarthomeNG/smarthome/wiki/Logging----Best-Practices).

Some more info [here](https://docs.python.org/3/howto/logging.html#logging-basic-tutorial)

For the level:
```python
import logging
DEBUGVERBOSE = logging.DEBUG - 1
logger.log(DEBUGVERBOSE , 'this is my verbose log message')
```

should work (not tested however)

msinn:
The official Python documentation says:
> Defining your own levels is possible, but should not be necessary, as the existing levels have been chosen on the basis of practical experience. 

Two sentences later they state that it's
> a very bad idea to define custom levels if you are developing a library.

ohinckel:
Even if it is bad practice, it is the best if you require custom log levels. I don't see any better solution for it.

And - if one requires this logging - it can be configured explicitely for the plugin instead globally to avoid confusion with different plugin/libraries/etc. using same log levels and producing log messages.

avdevice/plugin.yaml
```diff
@@ -3,23 +3,162 @@ plugin:
...
+        type: int
+        default: 0
+        description:
+            de: "Ein Wert zwischen 0 und 3, um das Verboselevel des Debugloggers zu definieren. Je höher, desto mehr Nachrichten werden geloggt."
+            en: "Value between 0 and 3 to define the verbose level of the debug logger. The higher the value, the more debug messages."
+  
 item_attributes:
-    # Definition of item attributes defined by this plugin
+    avdevice_zone[0-4]:
+        type: name[a-z]
```
ohinckel: @msinn, what about the "regex-matching-attribute-name" feature (as mentioned in #81)? Could it be useful for numbered attribute names with same settings?

Further what is the type `name[a-z]`?

onkelandy:
Any conclusion concerning custom log levels?
Might try the logger.log stuff then...

bmxp:
Go ahead and try if it helps. 

> Developing a library 
IMHO means that you should not create other loglevels when e.g. write another library for PyPI. It is ok for this internal little project I think.

onkelandy:
oki, updated everything and creating new PR

PR #83, opened 2017-10-22
added plugin lirc to send IR-commands via lircd

E3EAT:


lirc/README.md
```diff
@@ -0,0 +1,115 @@
...
+e.g. DVDLIVINGROOM_POWER ist set to 5  => lircd will send 5x POWER via IR
+     DVDLIVINGROOM_POWER ist set to 1  => lircd will send 1x POWER via IR
+
+
+```
+LIRC:
+    REMOTE_DVDLIVINGROOM:
+        DVDLIVINGROOM_POWER:
+            type: num
+            lirc_instance: "livingroom"
```
ohinckel: This should not be used. Instead you should use the instance atributes:
```yaml
item:
  type: num
    lirc_remote@livingroom: "..."
    lirc_key@livingroom: "..."
```

lirc/README.md
```diff
@@ -0,0 +1,115 @@
...
+You can configure the logger to see specific outputs of the plugin. That's helpfull when setting up the plugin. To see e.g. the debug-outputs of
+the plugin you have to configure the logging.yaml file. The name of logger for every instance is created by "plugins.lirc" + "_instancename" 
+e.g. plugins.lirc_livingroom or plugins.lirc_hifisystem
+
+
+```
+loggers:
+    plugins.lirc_livingroom:
+       level: DEBUG
+```
```
ohinckel: Sorry, but I'm not a friend of custom logging implementations. If we want to be able to configure logging for a given instance of a plugin, we should implement a common solution, so that all the other plugin can benefit from it.

E3EAT: OK. As long there's no general solution for instance specific logging I would remove the specific logger name but keep the prefix on every outpupt line that allows to identify which instance gave the output. (loggercmd)

E3EAT:
I have done the requested changes and commited them with my last 2 commits. Please let me know if there are further actions required.

lirc/plugin.yaml
```diff
@@ -0,0 +1,50 @@
+# Metadata for the Smart-Plugin
+plugin:
+    # Global plugin attributes
+    type: gateway                  # plugin type (gateway, interface, protocol, system, web)
+    description:
+        de: 'Sendet Kommandos and lircd das wiederum IR-Signale an Ger�te mit IR-Schnittstelle versendet'
```
ohinckel: Typo in "and", should be "an".

lirc/__init__.py
```diff
@@ -26,15 +26,15 @@
...
 from lib.model.smartplugin import SmartPlugin
 
-REMOTE_ATTRS = ['lirc_instance', 'lirc_remote', 'lirc_key']
+REMOTE_ATTRS = ['lirc_remote', 'lirc_key']
 class LIRC(lib.connection.Client,SmartPlugin):
 
     ALLOW_MULTIINSTANCE = True
     PLUGIN_VERSION = "1.3.1"
 
     def __init__(self, sh, lirc_Host, lirc_Port=8765):
```
ohinckel: Does it matter using uppercase in arguments? The `plugin.yaml` named them in lowercase.

lirc/plugin.yaml
```diff
@@ -0,0 +1,50 @@
...
+        description:
+            de: 'Name der Plugin Instanz (SmartHomeNG Standard-Parameter). Die Angabe ist notwendig um den passenden lircd zu ermitteln wenn ein Befehl versendet werden soll.'
+            en: 'Name of the plugin instance (SmartHomeNG standard parameter). This is mandatory to identify the matching lircd for a pending command.'
+
+    instance:
+        type: str
+        mandatory: True
+        description:
+            de: 'Eindeutiger Bezeichner der die Instanz benennt'
+            en: 'unique identifier that names the given instance'
```
ohinckel: Do we still need the `instance` parameter here?

ohinckel:
Only minor issues. It looks good to me.

One thing I dont't know how to deal with it, is the instance parameter. Usually this is a standard feature of the plugin handling but the docs here are telling the user really much about it, which is IMHO all described in the common plugin documentation (if not it should be there). But I think it is not a blocker, since more documentation is in most cases better than less. On the other hand changes on the instance handling could need to update the docs here too.

What do the other think about this PR?

psilo909:
@ohinckel i always like, if someone is able to test the pr. but if no one of the "normal" team can do that, we can accept it..

E3EAT:
I have fixed the typo and the duplicate description.
Regarding the instance parameter: I had the same thoughts on that but IMHO it is important to mention that the instance-name has to defined and used with all items that are used for the plugin.

msinn:
The plugin.yaml file is not in UTF-8 format. That broke the documentation build. 
I am going to update that file on github.

msinn:
I have changed the plugin.yaml. I commented out the documentation entry. It is for additional information, not to create a link to the README.md of the plugin. That link is known an automatically embedded in the shng documentation.

E3EAT:
OK, I'm sorry for the circumstances. I will have an eye on that in the future.

ohinckel:
> That broke the documentation build.

Can this be integrated into a Travis build so that we get notified about it? This could also update the developer docs automatically if someone commits.

msinn:
> Can this be integrated into a Travis build so that we get notified about it? This could also update the developer docs automatically if someone commits.

I know that building the documentation on every commit is your favorite. But before thinking of that, we need a complete redesign of the way tests are run for plugins:

At the moment, when you make a commit to a plugin no test is run at all. But if you make a commit to the core, the plugin tests are run.

If someone commits a faulty plugin an you make a commit to the core afterwards, the core build is flaged as faulty.

ohinckel:
I think we do not need to redesign the testing suite. We only need to configure Travis to run the tests in the right (tm) way. For example by checking out both repository in the develop version and then run the tests.

If we want to have the tests running for both (core and plugins) let me know. I will try to set this up. IMHO I would configure the following in Travis:
* each commit to master of core will just run the tests in master branch with plugin version of this master branch
* each commit to develop core will run the tests in develop branch with plugin version head of develop of plugins
* each commit to develop of plugin will run the tests like the one above (both with head of develop branches)

With this we do not need to bump the plugin repo version in core to the latest to usethe develop version. But we could also add a commit hook which updates the plugin version in core to the latest version each time one commits to the plugin repo. Will also result in executing tests with the latest versions.

msinn:
I opend issue https://github.com/smarthomeNG/smarthome/issues/240 for that discussion.

PR #86, opened 2017-10-30
MPD: added support for loading and playing local playlists

E3EAT:
added support for loading and playing local playlists which means sending the following commands to MPD to play a playlist:

clear
load filename
play

### DIFF
--- a/mpd/README.md
+++ b/mpd/README.md
@@ -37,6 +37,10 @@ You could see a full featured item configuration at the end of this file.
             type = bool
             mpd_file = http://jungletrain.net/64kbps.m3u
             enforce_updates = yes
+        [[[heavy]]]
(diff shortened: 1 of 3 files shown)


ohinckel:
Looks good to me. But we could also start updating the `plugin.yaml` and add `yaml` example to the `README.md`.

E3EAT:
updated as requested. Not sure if there are further changes required in the plugin.yaml.

ohinckel:
The parameters for the plugin and the items can be described in the `plugin.yaml`. But I think this can be added in one of the next PRs. Further it could be converted to be a [SmartPlugin](https://github.com/smarthomeNG/smarthome/wiki/SmartPlugin).

So far, thanks for your enhancement!

PR #93, opened 2017-11-05
Database plugin: more aggregation function sum, count, val

ohinckel:
This is the implementation for issue https://github.com/smarthomeNG/smarthome/issues/235 which includes new aggregation functions and introduce finalizer function.

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -138,6 +138,15 @@ Supported functions are:
    * `min`: for the minimum value
    * `on`: percentage (as float from 0.00 to 1.00) where the value has been greater than 0.
    * `sum`: for the summarized value
+   * `val`: for the raw values
+
+Additionally to the aggregation function a finalizer function can be specified when
+fetching series to apply to the results before returning them. Specify the function
+as prefix to the actual aggregation function (e.g. "diff:avg").
+
+Supported finalizer functions are:
+
+   * diff: return the differences between values
 
 For the timeframe you have to specify a start point and a optional end point. By default it ends 'now'.
 The time point could be specified with `<number><interval>`, where interval could be:
--- a/database/README.md
+++ b/database/README.md
@@ -133,7 +133,7 @@ for the specified function and timeframe.
 Supported functions are:
 
    * `avg`: for the average value
-   * `count`: for the count of value!=0 (more examples: `count>10`, `count<10`, `count=10`)
+   * `count`: for the amount of values not "0" (more examples: `count>10`, `count<10`, `count=10`)
    * `max`: for the maximum value
    * `min`: for the minimum value
    * `on`: percentage (as float from 0.00 to 1.00) where the value has been greater than 0.
@@ -146,7 +146,7 @@ as prefix to the actual aggregation function (e.g. "diff:avg").
 
 Supported finalizer functions are:
 
-   * diff: return the differences between values
+   * `diff`: return the differences between values
 
 For the timeframe you have to specify a start point and a optional end point. By default it ends 'now'.
 The time point could be specified with `<number><interval>`, where interval could be:
--- a/database/README.md
+++ b/database/README.md
@@ -133,9 +133,11 @@ for the specified function and timeframe.
 Supported functions are:
 
    * `avg`: for the average value
+   * `count`: for the count of value!=0 (more examples: `count>10`, `count<10`, `count=10`)
    * `max`: for the maximum value
    * `min`: for the minimum value
    * `on`: percentage (as float from 0.00 to 1.00) where the value has been greater than 0.
+   * `sum`: for the summarized value
 
 For the timeframe you have to specify a start point and a optional end point. By default it ends 'now'.
 The time point could be specified with `<number><interval>`, where interval could be:
--- a/database/README.md
+++ b/database/README.md
@@ -140,14 +140,6 @@ Supported functions are:
    * `sum`: for the summarized value
    * `val`: for the raw values
 
-Additionally to the aggregation function a finalizer function can be specified when
-fetching series to apply to the results before returning them. Specify the function
-as prefix to the actual aggregation function (e.g. "diff:avg").
-
-Supported finalizer functions are:
-
-   * `diff`: return the differences between values
-
 For the timeframe you have to specify a start point and a optional end point. By default it ends 'now'.
 The time point could be specified with `<number><interval>`, where interval could be:
 
@@ -174,6 +166,20 @@ e.g.
 sh.outside.temperature.series('min', '1d', count=10)  # returns 10 minimum values within the last day
 sh.outside.temperature.series('avg', '2w', '1w')  # returns the average values of the week before last week
 </pre>
+
+Additionally to the aggregation function a finalizer function can be specified when
+fetching series to apply to the results before returning them. Specify the function
+as prefix to the actual aggregation function (e.g. "diff:avg").
+
+Supported finalizer functions are:
+
+   * `diff`: return the differences between values
+
+e.g.
+<pre>
+sh.outside.temperature.series('diff:avg', '2w', '1w')  # returns the differences between average values
+</pre>
+
 
 ### sh.item.dbplugin
 This property returns the associated `database` plugin instance. See the list of method below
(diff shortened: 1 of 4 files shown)


psilo909:
@ohinckel do i understand count right? 

e.g. sh.eta_unit.temperature_outside.value.db('count>1', '24h','now') should return me the number of values in the last 24h that are larger than 1?

currently i only get None in each try i did with count

ohinckel:
This PR only supports the new functions in the `series` method. Don't know if it is useful in the `db` function too. Maybe I can implement this too.

> I dont get what you mean with "for the count of value!=0"

As mentioned above, it is only supported in the `series` method. And right, the documentation is not fully correct. I was assuming that I implement the "count!=0" as default. But that's wrong, there is no default. You always have to specify the condition. But a default (e.g. "!=0" or "<>0") could be useful, too.

> returns more than 10 values:

I think this was also the case in the previous version of the plugin and not introduced by this PR.  Could you provide the SQL statement which is executed and the raw values? Just to let me know if this is already returned by the SQL database or added in the code manually later.

psilo909:
Hi,
when it is only supported in the series method, why did u document it to the db method then? :-)

psilo909:
```
2017-11-08  10:08:25 ERROR    root             SELECT MIN(time), MIN(val_num) FROM log WHERE item_id = %(id)s AND time >= (SELEC
T COALESCE(MAX(time), 0) FROM log WHERE item_id = %(id)s AND time < %(time_start)s) AND time <= %(time_end)s AND time + COALESCE
(duration, %(inow)s - time) > (SELECT COALESCE(MAX(time), 0) FROM log WHERE item_id = %(id)s AND time < %(time_start)s) GROUP BY
 ROUND(time / %(step)s)                                                                                                         
2017-11-08  10:08:25 ERROR    logics.Test      {'sid': 'eta_unit.temperature_outside.value|min|1d|now|10', 'params': {'start': 1
510132105437, 'func': 'min', 'end': 'now', 'update': True, 'item': 'eta_unit.temperature_outside.value', 'sid': 'eta_unit.temper
ature_outside.value|min|1d|now|10', 'step': 8640000}, 'cmd': 'series', 'update': datetime.datetime(2017, 11, 8, 12, 32, 25, 5035
86, tzinfo=tzfile('/usr/share/zoneinfo/Europe/Berlin')), 'series': [(1510045705437, 6.0), (1510051797439, 6.0), (1510061129333, 
6.4), (1510069456505, 6.3), (1510079410311, 6.3), (1510088355234, 6.0), (1510095706400, 5.4), (1510103866410, 5.0), (15101129892
90, 2.9), (1510121210285, 2.4), (1510129490426, 4.8), (1510132025242, 5.6), (1510132105437, 5.6)]} 
```

ohinckel:
Because the list of supported functions are only in the documentation for `db()` and the `series()` method references this. This should be clarified, thanks for the hint. I would update it to support this for the `db()`  method too.

psilo909:
some more:
```
2017-11-08  10:16:26 ERROR    root             SELECT MIN(time), MIN(val_num) FROM log WHERE item_id = %(id)s AND time >= (SELECT COALESCE(MAX(time), 0) FROM log WHERE item_id = %(id)s AND time < %(time_start)s) AND time <= %(time_end)s AND time + COALESCE(duration, %(inow)s - time) > (SELECT COALESCE(MAX(time), 0) FROM log WHERE item_id = %(id)s AND time < %(time_start)s) GROUP BY ROUND(time / %(step)s) 

2017-11-08  10:16:26 ERROR    root             {'time_end': 1510132586440, 'step': 8640000, 'id': 5, 'inow': 1510132586440, 'time_start': 1510046186440}

2017-11-08  10:16:26 ERROR    logics.Test      {'update': datetime.datetime(2017, 11, 8, 12, 40, 26, 498570, tzinfo=tzfile('/usr/share/zoneinfo/Europe/Berlin')), 'params': {'start': 1510132586440, 'sid': 'eta_unit.temperature_outside.value|min|1d|now|10', 'end': 'now', 'update': True, 'func': 'min', 'step': 8640000, 'item': 'eta_unit.temperature_outside.value'}, 'cmd': 'series', 'sid': 'eta_unit.temperature_outside.value|min|1d|now|10', 'series': [(1510046186440, 6.0), (1510051797439, 6.0), (1510061129333, 6.4), (1510069456505, 6.3), (1510079410311, 6.3), (1510088355234, 6.0), (1510095706400, 5.4), (1510103866410, 5.0), (1510112989290, 2.9), (1510121210285, 2.4), (1510129490426, 4.8), (1510132513507, 5.8), (1510132586440, 5.8)]}
```

ohinckel:
The query returns more then $count item because:
* +1 item because the query includes the last value before $start too (was requested in #63)
* +1 item when the item's last update is between $start and $end to reflect recent item changes
* +1 item when $end is not the last timestamp from query by copying last entry with $end as timestamp

This means that the result could include up to 3 more items than requested.

psilo909:
@ohinckel what about the "Decimal" and float mixture in the array below? is this a problem?

logics.Test      {'update': datetime.datetime(2017, 11, 9, 18, 56, 18, 245285, tzinfo=tzfile('/usr/share/
zoneinfo/Europe/Berlin')), 'cmd': 'series', 'sid': 'utility_room.water_meter.trigger|count|1d|now|100', 'series': [(1510247926135, Deci
mal('0')), (1510249162132, Decimal('4')), (1510249185707, 0.0), (1510249314185, 0.0)], 'params': {'step': 864000, 'sid': 'utility_room.
water_meter.trigger|count|1d|now|100', 'end': 'now', 'func': 'count', 'update': True, 'item': 'utility_room.water_meter.trigger', 'star
t': 1510249314185}}

ohinckel:
I think the `sqlite` plugin does add items too. At least I used this implementation as starting point. Which problems exists with the minmaxavg chart? Maybe @smailee can provide an example.

The Decimal should not be a problem. The values are converted by the `json` module - whatever it does with such values - and I hope it does the right thing. The database plugin does not do any conversion of the result set returned by the DB-API layer and simply use the return values directy. Som of the layers are returning Decimal other return other types (e.g. int).

psilo909:
https://knx-user-forum.de/forum/supportforen/smartvisu/1103096-plot-minmaxavg-in-v2-8/page2

psilo909:
discussion about bar width

smaiLee:
Yes, I can.

The following call 
I call e.g. `{{ plot.period('', 'item', 'minmax', '1d', '', '', '', '10') }}` which generates `item.minmax.1d.now.10` and get 13 datapoints instead of the expected 10.

You see the distribution in time in [this fiddle](http://jsfiddle.net/jrppfhbw/1/) (I changed the x-values for better comprehensibility, but the time values are original).
The final 3 points are very close to each other, so highcharts shrinks the bar width.

smaiLee:
@ohinckel Sorry, I did not see your [last comment 2 days ago](https://github.com/smarthomeNG/plugins/pull/93#issuecomment-342943556).

The actual problem is not the 3 additional entries but the density of them. As mentioned this leads to the slim bars.

Foxi352:
I think i don't get how this is supposed to work. I have a boolean in my db going to True when the flame in my boiler starts, and to False when the flame stops. In Smartvisu i want a graph showing the last 24h with one bar per hour showing the number of flame starts for that hour.

What i tried (but does not work):
```
{{ plot.period('', 
   'buderus.waermeerzeuger.flamme', 
   'count', '24h', '', '', '', '',
  '',
  '', 'column',
  '') }}  
```
Is what i want even possible ? To be honest: That was the point of my feature request ;-)

smaiLee:
If you want one bar per hour in the last 24 hours you have to set 8th argument to 24:
```
{{ plot.period('', 
   'buderus.waermeerzeuger.flamme', 
   'count', '24h', '', '', '', '24',
  '',
  '', 'column',
  '') }}  
```

But as currently discussed on [Gitter](https://gitter.im/sVISU/Lobby?at=5a0578d8df09362e67ef2472) then you get 3 more values which don't make sense in this use case.

psilo909:
@Foxi352 can we merge this PR? It wont fix the mix-max-avg Chart issue, but it also did not bringt it in. for what the PR was aimed at is done correctly...
Do you have any open points?

Foxi352:
@psilo909 absolutely ... No objection from my side.

PR #96, opened 2017-11-16
Avdevice plugin: major update

onkelandy:
optimized lots of functions
made compatible with new plugin.yaml while keeping it compatible to 1.3 master (hopefully)
removed statusupdate from device txt file as this function is implemented in the plugin already.


ohinckel:
Do we need to mention this change in the 1.4 release wiki page?

psilo909:
i doubt the plugin was part of 1.3 ?!

psilo909:
uhh it was.. but i think it was never really working.. the last update is the first time it is really working at my place

onkelandy:
It was fully working for Pioneer receivers and supported limited features for Denon back then.

ohinckel:
@psilo909 it doesn't matter if it was part of 1.3 or not. I only think we should include all the changes/features for a given release in the release notes in wiki.

@onkelandy maybe you can add some notes to the (wiki)[https://github.com/smarthomeNG/smarthome/wiki/SmartHomeNG-1.4]?

onkelandy:
done

psilo909:
i think i already added a line before @onkelandy did.. i suppose we have that twice now

psilo909:
it looks like my revision (https://github.com/smarthomeNG/smarthome/wiki/SmartHomeNG-1.4/2088c789dbebb99a6b807c99817317249c927f7d) was overwritten.. we should at least place a statement that the txt-file definition and the items have changed.

PR #98, opened 2017-11-20
MPD: major update: plugin is now a SmartPlugin; Added missing playpau…

E3EAT:
…se, mute, random, ... items; Added miscellaneous commands and status requests for a comfortable control of the playback; Updated README.md and plugin.yaml; Added database-commands;

I'm using this plugin daily and so far it's working stable.

### DIFF
--- a/mpd/README.md
+++ b/mpd/README.md
@@ -3,296 +3,518 @@
 ## Requirements
 
 You only need one or more Music Player Daemons (MPD).
+For more information about MPD it's commands see http://www.musicpd.org/doc/protocol/command_reference.html
 
 ## Configuration
 
-### plugin.conf (deprecated) / plugin.yaml
-
-```
-[mpd]
-    class_name = MPD
-    class_path = plugins.mpd
-```
+### plugin.yaml
 
 ```yaml
 mpd:
     class_name: MPD
     class_path: plugins.mpd
+    instance: "bathroom"
+    host: "192.168.177.164"
+    port: 6601
+#    cycle: 2
+
+mpd:
+    class_name: MPD
+    class_path: plugins.mpd
+    instance: "kitchen"
+    host: "192.168.177.164"
+    port: 6602
+#    cycle: 2
 ```
 
-### items.conf (deprecated) / items.yaml
+#### instance
+This attribute is mandatory. You have to provide an unique identifier for every instance.
 
-You could see a full featured item configuration at the end of this file.
+#### host
+This attribute is mandatory. You have to provide the IP adress or the hostname of the MPD.
 
+#### port
+You can specify a port to connect to. If not defined port 6600 will be used.
+
+#### cycle
+You can specify the update interval. If not defined cycle will be set to 2.
+
+
+###Command groups and commands
 ```
-[living]
-    type = bool
-    [[mpd]]
-        type = str
-        mpd_host = 127.0.0.1
-        mpd_port = 6600
-        [[[state]]]
-            type = str
-            mpd_listen = state
-            mpd_send = value  # sends the item value. e.g. sh.dev.mpd.state('play') will send 'play'
-        [[[volume]]]
-            type = num
-            mpd_listen = volume
-            mpd_send = volume
-        [[[jungle]]]
-            type = bool
-            mpd_file = http://jungletrain.net/64kbps.m3u
-            enforce_updates = yes
-        [[[heavy]]]
-            type = bool
-            mpd_localplaylist = heavymetal # will play the playlist heavymetal.m3u in the MPD's playlist directory
-            enforce_updates = yes
+There are 7 groups of commands tat can be used. Here's an overview of the available groups and their commands and datatypes
+
+mpd_status
+    Implements commands that are used to query MPD's status.
+
+    playpause        # bool, 0 or 1, 1 while music is played, also used as mpd_command
+    mute             # bool, 0 or 1, 1 when volume is set to 0%, also used as mpd_command
+    volume           # num,  0-100 or -1 if the volume cannot be determined, also used as mpd_command
+    repeat           # bool, 0 or 1, also used as mpd_command
+    random           # bool, 0 or 1, also used as mpd_command
+    single           # bool, 0 or 1, also used as mpd_command
+    consume          # bool, 0 or 1, also used as mpd_command
+    playlist         # num,  31-bit unsigned integer, the playlist version number
+    playlistlength   # num,  integer, the length of the playlist
+    mixrampdb        # num,  mixramp threshold in dB
+    state            # str,  play, stop, or pause
+    song             # num,  playlist song number of the current song stopped on or playing
+    songid           # num,  playlist songid of the current song stopped on or playing
+    time             # str,  total time elapsed (of current playing/paused song)
+    elapsed          # str,  Total time elapsed within the current song, but with higher resolution
+    bitrate          # num,  instantaneous bitrate in kbps
+    audio            # str,  The format emitted by the decoder plugin during playback, format: "samplerate:bits:channels"
+    nextsong         # num,  playlist song number of the next song to be played
+    nextsongid       # num,  playlist songid of the next song to be played
+    duration         # num,  Duration of the current song in seconds
+    xfade            # num,  crossfade in seconds
+    mixrampdelay     # num,  mixrampdelay in seconds
+    updating_db      # str,  job id
+    error            # str,  errormessage of MPD
+
+mpd_songinfo
+    Implements commands that are used to query information about the current playing song.
+
+    file             # str,  path of the current playing file
+    Last-Modified    # str,  Date of last modification of the song
+    Artist           # str,  Artist of the current playing song
+    Album            # str,  Album of the current playing song
+    Title            # str,  Title of the current playing song
+    Track            # str,  Track# of the current playing song
+    Time             # str,  length of the current song
+    Pos              # str,  position of the song in the playlist
+    Id               # str,  Id of the song
+
+mpd_statistic
+    Implements commands that are used to query MPD's statistics.
+
+    artists          # num,  number of artists
+    albums           # num,  number of albums
+    songs            # num,  number of songs
+    uptime           # num,  daemon uptime in seconds
+    db_playtime      # num,  sum of all song times in the db
+    db_update        # num,  last db update in UNIX time
+    playtime         # num,  time length of music played
+
+mpd_command
+    Implements commands that are used to control the playback.
+
+    playpause        # bool, 0 or 1, 1 to play music, also used as mpd_status
+    mute             # bool, 0 or 1, 1 to set the volume to 0%, also used as mpd_status
+    volume           # num,  0-100 to set the volume, also used as mpd_status
+    repeat           # bool, 0 or 1, also used as mpd_status
+    random           # bool, 0 or 1, also used as mpd_status
+    single           # bool, Toggles single state. When single is activated, playback is stopped after current
+                     #       song, or song is repeated if the 'repeat' mode is enabled, also used as mpd_status
+    consume          # bool, Sets consume state to STATE, STATE should be 0 or 1. When consume is activated,
+                     #       each song played is removed from playlist, also used as mpd_status
+
+    next             # bool, Plays next song in the playlist
+    pause            # bool, Toggles pause/resumes playing, PAUSE is 0 or 1
+    play             # num,  Begins playing the playlist at song number <value>
+    playid           # num,  Begins playing the playlist at song <value>
+    previous         # bool, Plays previous song in the playlist
+    seek             # str,  Seeks to the position TIME (in seconds; fractions allowed)
+                     #       of entry SONGPOS in the playlist format: "SONGPOS TIME"
+    seekid           # str,  Seeks to the position TIME (in seconds; fractions allowed) of song SONGID
+                     #       format: "SONGID TIME"
+    seekcur          # str,  Seeks to the position TIME (in seconds; fractions allowed) within the current song
+                     #       If prefixed by '+' or '-', then the time is relative to the current playing position
+    stop             # bool, Stops playing
+    crossfade        # num,  Sets crossfading between songs
+    mixrampdb        # num,  Sets the threshold at which songs will be overlapped. Like crossfading but doesn't 
+                     #       fade the track volume, just overlaps
+    mixrampdelay     # num,  Additional time subtracted from the overlap calculated by mixrampdb
+    setvol           # num,  Sets volume to VOL, the range of volume is 0-100
+    replay_gain_mode # str,  ets the replay gain mode. One of off, track, album, auto
+
+mpd_url
+    Parses the defined url and tries to play the stream.
+
+mpd_localplaylist
+    Searches for a local playlist with the defined filename and plays it if found.
+
+mpd_rawcommand
+    Takes whatever you assign to it and sends it directly to MPD.
+
+mpd_database
+    Implements commands that are used to control the database.
+
+    update           # str,  Updates the music database: find new files, remove deleted files, 
+                     #       update modified files. <value> is a particular directory or song/file to update
+                     #       If you do not specify it, everything is updated
+    rescan           # str,  Same as update, but also rescans unmodified files
 ```
+
+For some items it is useful to combine the status and the matching command in one item, e.g. playpause, mute, volume,
+repeat, random, consume,...
+See the items.yaml for examples.
+
+
+###items.yaml
+
+1st example shows all items
 
 ```yaml
-living:
-    type: bool
-    mpd:
-        type: str
-        mpd_host: 127.0.0.1
-        mpd_port: 6600
-        state:
-            type: str
-            mpd_listen: state
-            mpd_send: value  # sends the item value. e.g. sh.dev.mpd.state('play') will send 'play'
-        volume:
-            type: num
-            mpd_listen: volume
-            mpd_send: volume
-        jungle:
-            type: bool
-            mpd_file: http://jungletrain.net/64kbps.m3u
-            enforce_updates: yes
-        heavy:
-            type: bool
-            mpd_localplaylist: heavymetal # will play the playlist heavymetal.m3u in the MPD's playlist directory
-            enforce_updates: yes
+EG:
+    BATHROMM:
+        MUSIC:
+            volume:
+                type: num
+                mpd_status@bathroom: volume
+                mpd_command@bathroom: setvol
+                enforce_updates: true
+
+            repeat:
+                type: bool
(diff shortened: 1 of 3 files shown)


mpd/README.md
```diff
@@ -3,296 +3,518 @@
...
+    port: 6601
+#    cycle: 2
+
+mpd:
+    class_name: MPD
+    class_path: plugins.mpd
+    instance: "kitchen"
+    host: "192.168.177.164"
+    port: 6602
+#    cycle: 2
```
ohinckel: Usually you don't need to quote all the strings in the YAML configuration files.

mpd/README.md
```diff
@@ -3,296 +3,518 @@
...
 ```
 
-### items.conf (deprecated) / items.yaml
+#### instance
+This attribute is mandatory. You have to provide an unique identifier for every instance.
```
ohinckel: Is this really useful? In other setup you register the plugin multiple times using the `instance` setting (as mentioned above) and use the settings like `<attribute-name>@<instance-name> = <value>`.

Can this `instance` attribute for an item be replaced with the syntax `<attrib>@<inst>` and using mjulti-instance plugin?

ohinckel: Just saw this in the YAML example below: `mpd_status@bathroom: volume` - so seems to be OK then.

mpd/README.md
```diff
@@ -3,296 +3,518 @@
...
-            mpd_listen = name
-        [[[track]]]
-            type = str
-            mpd_listen = track
+
+## Logging
+
+You can configure the logger to see specific outputs of the plugin. That's helpfull when setting up the plugin. To see e.g. the debug-outputs of
+the plugin you have to configure the logging.yaml file. 
+
```
ohinckel: The logging should be something generic and does not need to be mentioned here. We should describe the logging in a more generic way in the documentation and I guess this is already [there](http://smarthomeng.de/v1_3/developer/logging.html).

ohinckel:
Sorry for the late feedback. Overall status looks OK, so a merge would be OK too. But I can't test all the stuff. Maybe somebody can test it before we merge it?

mpd/README.md
E3EAT: Are there any exceptions? Some situations where quoting is necessary? I started quoting strings because I ran into problems with an url that was was not quoted in a YAML file.

ohinckel: In most cases I don't quote something. Only in special cases where values contains special characters (like ":", "=", "\r", "\n", ..) or in situations producing problems reading the value.

But it's not a big deal to quote it here in the example. It may confuse users seeing some examples with quotes, other without.

E3EAT: Ok, that clears up. I will remove the quotes and have a look to your ohter comments during the next week.

E3EAT:
I have just pushed the requested changes.
There are no new testers until now.

mpd/README.md
```diff
@@ -3,296 +3,506 @@
...
-```
+### plugin.yaml
 
 ```yaml
 mpd:
     class_name: MPD
     class_path: plugins.mpd
+    instance: bathroom
+    host: 192.168.177.164
+    host: 192.168.177.164
```
ohinckel: Duplicate `host` attribute necessary?

In case it is always an IP address the `plugin.yml` can be changed to declare this field as a `ip` field instead of `str`. But I guess it could also be just a hostname.

E3EAT: Yes, I just testet using the hostname instead of the IP in the plugin.yaml. Works as expected with a hostname too.

mpd/plugin.yaml
```diff
@@ -5,21 +5,47 @@ plugin:
...
+    version: 1.3.1                 # Plugin version
+    sh_minversion: 1.3             # minimum shNG version to use this plugin
 #    sh_maxversion:                 # maximum shNG version to use this plugin (leave empty if latest)
-#    multi_instance: False
+    multi_instance: True
     classname: MPD                 # class containing the plugin
 
 #parameters:
     # Definition of parameters to be configured in etc/plugin.yaml
-
```
ohinckel: I guess the following parameters are plugin parameters. So you need to uncomment the `parameters` section here.

mpd/README.md
ohinckel: OK, thanks for testing this. But I meant to change the `type` of the `host` parameter in the `plugin.yml` from `str` (current version) to `ip`. But I guess this is not possible because `ip` type is matching against a valid IP string. So hostnames are not valid for `ip` type. @msinn can tell us more about the parameter validation.

For the moment it's OK as it is currently.

msinn: At the moment there are two datatypes that can be used for hostnames:
- ip - a string, representing a hostname or an ipv4-address
- ipv4 - a string, representing strictly an ipv4-address

Metadata is only documented in the development documentation of the (not yet released) v1.4: https://www.smarthomeng.de/dev/developer/development_plugin/plugin_metadata.html

E3EAT:
Yes you're right. Fixed with my last commit

ohinckel:
Anyone able to test it before we merge?

E3EAT:
I have not tested the plugin with the new release of SHNG yet so I don't know if there are changes required for the new release. I will test that asap. Shall I add potential commits with adjustments to this PR if necessary?

ohinckel:
At least a test run with the new SHNG version (or develop) would be great. If there are some problems, just fix them and add the commits to this PR.

morphZ:
Hi Everybody. I'm relying heavily on the mpd plugin and have the following problem: If I send commands by other clients (e.g. mpc or MPdroid) to the mpd server, that smarthomeNG is connected to via mpd-plugin, the plugin gets unresponsive. I.e. play, stop and state attributes are no longer updated. Is that a general problem or something in my personal setup. Glad for any hints. Thanks!

BTW: Thanks for your work on this great project. If I should create an issue on the main project for that just give me hint. I'd even try to fix it myself, if you give me some place to start.

E3EAT:
just committed some adjustments for SHNG 1.4.

@morphZ 
I'm also using MPDroid simultaniously with the SHNG plugin but I can't reproduce the behaviour you described. 
But: If your cycle time in the plugin.yaml is too high you can get some "overrides" between your client and the plugin. So a good start is to reduce the cycle time. Default time is 2 seconds and that's fine if you do not work with another client simultaniously.

What are your versions of SHNG, MPD-Plugin, MPDroid and MPD-Server? Is a restart of the plugin required after it became unresponsive? Do you want to act as additional tester for the plugin?
I'm not sure if creating an Issue at this stage is the best way because the rewrite of the plugin is currently only in the development-branch and still with this open PR.

@ohinckel 
What do you think? Where's the best place for communications to figure out what causes morphZ's problems?

ohinckel:
@E3EAT, I would suggest:
* if the problems are related to the plugin code of this PR then just comment it here
* if the problems are within the current version in (`master` / `develop`) then create a new issue

morphZ:
@E3EAT 
I'm using SHNG and plugins from the master branch of the main repo. Upgraded just a few days ago.

#### module versions
SHNG 1.4.1
MPD 1.3.3 (according to plugin.yaml)
MPdroid 1.07.2 Final ba27d2a-c (54)
mpd 0.19.21

Note: It's probably not a problem of MPdroid, as I get the same behavior if I issue commands via mpc in the shell.

A restart of mpd fixes the problem. A restart of SHNG fixes the problem. How can I restart a single plugin?

What about if I switch to your PR version and try that instead? We can focus on getting that PR merged instead of fixing some soon deprecated code. I can try to help as tester. How can I help? Write some unit tests?

morphZ:
OK, here's some feedback:

I use a very slim configuration to control mpd via a KNX-Controller:

    # items/11_mpd.yaml

    Musik:
        volume:
            type: num
            enforce_updates: true
            mpd_status@wohnen: volume
            mpd_command@wohnen: setvol
        play:
            type: bool
            enforce_updates: true
            mpd_status@wohnen: playpause
            mpd_command@wohnen: playpause

One thing is: I get errors from the `update_currentsong()` and `update_statistic()` methods just because I don't have corresponding items in my config. Shouldn't that be INFO-level at most?

morphZ:
I have a similar issue as with the old plugin: The status is not being parsed. This looks like this in the log:

    2018-01-07  23:02:53 DEBUG    plugins.mpd      MPD_wohnen: requesting status
    2018-01-07  23:02:53 DEBUG    plugins.mpd      MPD_wohnen: send status to MPD
    2018-01-07  23:02:53 DEBUG    lib.scheduler    MPD_wohnen next time: 2018-01-07 23:02:55+01:00

After restarting mpd via systemctl I get the usual:

    2018-01-07  23:05:19 DEBUG    lib.scheduler    MPD_wohnen next time: 2018-01-07 23:05:21+01:00
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: requesting status
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: send status to MPD
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => volume: 4
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => repeat: 1
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => random: 0
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => single: 0
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => consume: 0
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => playlist: 4
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => playlistlength: 4
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => mixrampdb: 0.000000
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => state: play
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => song: 0
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => songid: 1
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => time: 4:0
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => elapsed: 4.498
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => bitrate: 128
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => audio: 48000:24:2
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => nextsong: 1
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => nextsongid: 2
    2018-01-07  23:05:21 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => OK

Here is my plugin.yaml snippet:

    mpd:
        class_name: MPD
        class_path: plugins.mpd
        instance: wohnen
        host: localhost
        port: 6600

morphZ:
Another thing: *Every* cycle the `update_status()` somehow "changes" the volume, triggering the `on_change()` attribut of smarthome. Event though the volume has not changed at all. Here is the log:

    2018-01-07  23:23:04 DEBUG    plugins.mpd      MPD_wohnen: update item volume with 20
    2018-01-07  23:23:04 INFO     lib.item         Item Musik.volume: 'on_change' evaluating Musik.cmd_info.volume = value
    2018-01-07  23:23:04 DEBUG    lib.item          - : 'on_change' finally evaluating Musik.cmd_info.volume = value, result=20.0

morphZ:
Next thing: When I try to start mpd (`state=='stop' at the time`) it sends the command `pause 0` when it should send `play` to mpd.

    2018-01-08  09:10:59 DEBUG    lib.item         Item Musik.play = True via Logic None None
    2018-01-08  09:10:59 DEBUG    plugins.mpd      MPD_wohnen: send pause 0 to MPD
    2018-01-08  09:10:59 DEBUG    plugins.mpd      MPD_wohnen: parse_reply => OK

morphZ:
Will there be a `.command(cmd)` method like in the old plugin to send commands to mpd directly? I cannot call the `._send()` method from a logic.

E3EAT:
Thanks for yourfeedback it's very appreciated! This is what I found out so far:

> One thing is: I get errors from the update_currentsong() and update_statistic() methods just because I don't have corresponding items in my config. Shouldn't that be INFO-level at most?

I guess you mean the following output?
currentsong: no items to refresh
status: no items to refresh

These messages are warnings and not errors and will be printed only once per hour. IMHO that messages are type warning and not info.

            
           
> Another thing: Every cycle the update_status() somehow "changes" the volume, triggering the on_change() attribut of smarthome. Event though the volume has not changed at all. Here is the log:

Yes that was a bug and I fixed that with the last commit. Please update your plugin.

> Next thing: When I try to start mpd (state=='stop' at the time) it sends the command pause 0 when it should send play to mpd.

I also implemeted that with the last commit. Please update your plugin.

> Will there be a .command(cmd) method like in the old plugin to send commands to mpd directly? I cannot call the ._send() method from a logic.

I'm not familiar with the interaction of plugins and logics. But mpd_rawcommand in the plugin should fit your needs. Please take a look to the README.md: 
mpd_rawcommand
    Takes whatever you assign to it and sends it directly to MPD.

> I have a similar issue as with the old plugin: The status is not being parsed. This looks like this in the log:
2018-01-07  23:02:53 DEBUG    plugins.mpd      MPD_wohnen: requesting status
2018-01-07  23:02:53 DEBUG    plugins.mpd      MPD_wohnen: send status to MPD
2018-01-07  23:02:53 DEBUG    lib.scheduler    MPD_wohnen next time: 2018-01-07 23:02:55+01:00
EDIT: the mpd log seems no different in that case. It logs (every 2 secs):
Jan 08 09:30 : client: [37] process command "status"
Jan 08 09:30 : client: [37] command returned 0


I have to do further investigations in that. I definitly do not have that behaviour when I control MPD with other clients. On the first view I would suppose that MPD changes the terminator and therefore the plugin doesn't read the response. I will try to find out more about that and let you know.

mpd/__init__.py
```diff
@@ -21,120 +22,381 @@
...
-        if play != []:
-            self._send("add {0}".format(play[0]), False)
+    def update_statistic(self,warn):
+        if not self.connected:
+            if warn:
+                self.loggercmd("update_statistic while not connected",'e')
+            return
+        if (len(self._statistic_items) <= 0):
+            if warn:
+                self.loggercmd("statistic: no items to refresh",'e')
```
morphZ: Shouldn't it be just a warning here?

mpd/__init__.py
```diff
@@ -21,120 +22,381 @@
...
-        self._send('load ' + file, False)
-        self._send('play', False)
+    def update_currentsong(self,warn):
+        if not self.connected:
+            if warn:
+                self.loggercmd("update_currentsong while not connected",'e')
+            return
+        if (len(self._currentsong_items) <= 0):
+            if warn:
+                self.loggercmd("currentsong: no items to refresh",'e')
```
morphZ: Shouldn't it be just a warning here?

morphZ:
Hi! Thanks for the quick reply.

> These messages are warnings and not errors and will be printed only once per hour. IMHO that messages are type warning and not info.

I just added comments to the source at the places that I meant. Can you see them?

mpd/__init__.py
E3EAT: Yes I can see them and of course you are right. I have to be blind... I was sure they are warnings. Will harmonize them to warnings with the next commit.

morphZ:
> I have to do further investigations in that. I definitly do not have that behaviour when I control MPD with other clients. On the first view I would suppose that MPD changes the terminator and therefore the plugin doesn't read the response. I will try to find out more about that and let you know.

I sometimes get the problems right at the startup of shng. Without issuing any commands to mpd through other clients. I'm afraid, I cannot consistently reproduce that. Maybe some kind of race condition in the networking code?!

mpd/__init__.py
```diff
@@ -21,120 +22,381 @@
...
+                    except:
+                        self.loggercmd("can't parse {} to float".format(val),'e')
+                        continue
+                if item() != val:
+                    self.loggercmd("update item {} with {}".format(key,val),'d')
+                    self.setItemValue(item,val)
+            #update subscribed items which do not exist in the response from MPD
+            elif key == 'playpause':
+                item = subscribedItems[key]
+                self.setItemValue(item,self._internal_tems['isPlaying'])
```
morphZ: I get the same problem as with `volume` here (item `playpause` is update in every cycle). How about checking the same way as in the lines above (219 ff):

            elif key == 'playpause':
                item = subscribedItems[key]
                val = self._internal_tems['isPlaying']
                if item() != val:
                    self.setItemValue(item,val)

morphZ: Update: I changed that to a PR into your repo. Please try if it get's included into the PR to the main repo if you merge my Sub-PR. Just curious :)

morphZ:
Hi. I tested the `mute` command yesterday. Muting works, volume is set to zero. Unmuting works not, the volume just stays zero.

E3EAT:
> I sometimes get the problems right at the startup of shng. Without issuing any commands to mpd through other clients. I'm afraid, I cannot consistently reproduce that. Maybe some kind of race condition in the networking code?!

just spent some hours in testing and I also ran into that behaviour. I was able to find out what happend. MPD should always send "OK MPD (version)" when the connection is established. But sometimes MPD does not send that. In that case the terminator was not changed and the plugin was not able to parse any response. I made a change in the plugin, the terminator will change even when MPD does not send the inital handshake. Please update your plugin.

> Hi. I tested the mute command yesterday. Muting works, volume is set to zero. Unmuting works not, the volume just stays zero.

Yes, there was a missing parsing of a value. That's fixed now. Please update your plugin.

morphZ:
Great job! Thanks alot. This was actually quite an annoying bug, as it was hard to consistently reproduce.

How much more testing is to be done, until the PR has the chance to be merged?
Would you like me to test something specifically?

E3EAT:
from my side I would say we use the plugin another 1 or 2 weeks as much as possible and if no more issues occur I will ask ohinckel for a review/merge.

E3EAT:
the plugin ran stable and without any new Issues the last 17 days. I think that could be a good time for a review/merge.

@ohinckel, it would be nice if you could do that when you find the time.

morphZ:
Same here. Stable daily operation. No problems.

E3EAT <notifications@github.com> schrieb am Mo., 29. Jan. 2018, 17:24:

PR #101, opened 2017-12-09
Enhance Enocean Learn Method

hasenradball:
Enhance learn method with the following properties:
- one learn method for all actuators
- define different devices bei if...elif construction

### DIFF
--- a/enocean/README.md
+++ b/enocean/README.md
@@ -28,11 +28,11 @@ With the specification of the BaseID, 128 different transmit IDs are available, 
 
 ### Getting ID of an EnOcean device
 
-1.) reboot the pi or restart the smarthome (sudo reboot; sudo systemctl restart smarthome)
-2.) wait some time for comming up of the service
-3.) have a look into the log file an look for ``enocean: Base ID = 0xYYYYZZZZ``
-4.) now you have the right BaseID and you can place it into the plugin.conf-first
-5.) alternating you will also find the ChipID in the log-file
+1. reboot the pi or restart the smarthome (sudo reboot; sudo systemctl restart smarthome)
+2. wait some time for comming up of the service
+3. have a look into the log file an look for ``enocean: Base ID = 0xYYYYZZZZ``
+4. now you have the right BaseID and you can place it into the plugin.conf-first
+5. alternating you will also find the ChipID in the log-file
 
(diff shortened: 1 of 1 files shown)


hasenradball:
Change ist tested and works well.

ohinckel:
Once again, your diff looks very large because of changed line endings. Please check your settings how git/editor/... is dealing with crlf/lf.

I don't see what was actually changed in the `README.md` file, but the examples should also contain an example `yaml` configuration (for plugin and item configuration). Do you think you can add this too?

hasenradball:
Hello,

My changes was only in the area learn methods. So the changes should be not so much.
Just having placed the learn methods in a if... elif construction as we have discussed in the last pr.

In the Readme i have also changed the section with the learn methods.
 What do you mean with the line endings?
Is it necessary to adapt my editor?
I use the notepad++.

Yes i can also prepare a yaml example, but a smaller one.

Mit freundlichen Grüßen

Frank Häfele

hasenradball:
Hello ohinckel,

can you have a look in the rich diff then you see the real canges easily.
Do you have an hint for me which editor or which setting I sshould use to make the diffs easy to show?

ohinckel:
I know how to see the real diff, but I like to see it on github too (makes it easier to review and comment, and makes it easier for other not familar with other VCS tools).

With line endings I mean the characters used to indicate a line break. Different OSes use [different endings](https://en.m.wikipedia.org/wiki/Newline), which can be configured in the git client. Usually it's also be possible do configure this in your editor.

hasenradball:
Hi, I checked the line endings and for my understanding it is configured right.
which editor to you use?

How should we proceed?
Can you merge the pull request and I will prepare the yaml example and make an update of the readme in an few days?

hasenradball:
My line ending is set to autocrlf = true

ohinckel:
On Windows systems (using CRLF) `autocrlf = true` should be OK (means GIT will convert the LF to CRLF when checking out files and CRLF to LF back again when adding files to the index).

Currently the file does not use CRLF, instead it only uses LF. But with your setup I don't understand why the line endings were changed then.

Btw. if you want to do futher modifications to the files, simply commit to your branch and push it again. This will update your PR automatically including your new commits.

hasenradball:
closing and starting an new one.

PR #108, opened 2017-12-20
telegram: individual chat_id's, picture, attribute: 'telegram_info', list logoc, trigger logic

ivan73:
**new optional function-parameter to send message to individual chat_id's:**
    def _msg_broadcast(self, msg, chat_id=None):
**new function to send pictures**
    def _photo_broadcast(self, photofile, msg, chat_id=None):
**new SH-Item-attribute to read specific SH-items via bot with the /info - command**
	the plugin create a bot-keyboard.
    e.g. telegram_info = wetter # SH-Item-attribute
**new command /lo:**
	list all SH-logics
**new command /tr:**
	trigger SH-logic e.g. /tr garagentor_auf

### DIFF
--- a/telegram/README.md
+++ b/telegram/README.md
@@ -115,6 +115,7 @@ All attribute parameters (commands) are listed with the /info-command in a keybo
 
 #### Simple Example
 
+my_item_config.yaml
 ```
 [Aussentemperatur]
 	name = Aussentemperatur in °C
@@ -134,7 +135,7 @@ All attribute parameters (commands) are listed with the /info-command in a keybo
 	telegram_info = "rtr_ist"
 
 ```
-
+my_item_config.conf
 ```
 Aussentemperatur:
     name: Aussentemperatur in °C
--- a/telegram/README.md
+++ b/telegram/README.md
@@ -109,7 +109,8 @@ state_name:
 #### telegram_info
 
 read (broadcast) a list with specific item-values provided with the attribute.
-The attribute parameter is also the command. e.g. /wetter
+The attribute parameter is also the command. 
+e.g. /wetter
 All attribute parameters (commands) are listed with the /info-command in a keyboard menu
 
 #### Simple Example
@@ -141,21 +142,21 @@ Aussentemperatur:
     knx_dpt: 9
     telegram_info: wetter
 Wind_kmh:
-	name: Wingeschwindigkeit in kmh
-	type: num
-	knx_dpt: 9
-	telegram_info: "wetter"
+    name: Wingeschwindigkeit in kmh
+    type: num
+    knx_dpt: 9
+    telegram_info: "wetter"
 
 
 Raumtemperatur
-	name: Raumtemperatur Wohnzimmer in °C
-	type: num
-	knx_dpt: 9
-	telegram_info: rtr_ist
+    name: Raumtemperatur Wohnzimmer in °C
+    type: num
+    knx_dpt: 9
+    telegram_info: rtr_ist
 
 ```
-/info broadcast all info-commands in a bot-keyboard-menu
-    e.g.
+/info broadcast all info-commands in a bot-keyboard-menu e.g.
+    
     [ /wetter] [/rtr_ist]
 
 /wetter broadcast all items and values provided with the attribute 'telegram_info = "wetter"'
--- a/telegram/README.md
+++ b/telegram/README.md
@@ -106,6 +106,71 @@ state_name:
     telegram_message: 'New AutoBlind state: [VALUE]'
 ```
 
+#### telegram_info
+
+read (broadcast) a list with specific item-values provided with the attribute.
+The attribute parameter is also the command. e.g. /wetter
+All attribute parameters (commands) are listed with the /info-command in a keyboard menu
+
+#### Simple Example
+
+```
+[Aussentemperatur]
+	name = Aussentemperatur in °C
+	type = num
+	knx_dpt = 9
+	telegram_info = "wetter"
+[Wind_kmh]
+	name = Wingeschwindigkeit in kmh
+	type = num
+	knx_dpt = 9
+	telegram_info = "wetter"
+
+[Raumtemperatur]
+	name = Raumtemperatur Wohnzimmer in °C
+	type = num
+	knx_dpt = 9
+	telegram_info = "rtr_ist"
+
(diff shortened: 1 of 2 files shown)


telegram/README.md
```diff
@@ -106,6 +106,72 @@ state_name:
...
+    type: num
+    knx_dpt: 9
+    telegram_info: "wetter"
+
+
+Raumtemperatur
+    name: Raumtemperatur Wohnzimmer in °C
+    type: num
+    knx_dpt: 9
+    telegram_info: rtr_ist
```
ohinckel: Are all these quotes required in the configuration (e.g. `telegram_info: "wetter"`)?

telegram/__init__.py
```diff
@@ -160,40 +172,109 @@ def update_item(self, item, caller=None, source=None, dest=None):
...
+                
+    def _photo_broadcast(self, photofile, msg, chat_id=None):
+        chat_ids_to_send = []
+        if chat_id is None:
+            chat_ids_to_send = self._chat_ids
+        else:
+            if isinstance(chat_id, list):
+                chat_ids_to_send = chat_id
+            else:
+                chat_ids_to_send.append(chat_id)
```
ohinckel: Would it be possible to extract this part of code (checking for chat_id and constructing a list, line 191-198) into an own method and use it also at line 177?

telegram/__init__.py
```diff
@@ -160,40 +172,109 @@ def update_item(self, item, caller=None, source=None, dest=None):
...
+            #self._bot.sendMessage(self._chat_id, "Infos from the items:", reply_markup=self.create_info_reply_markup())
+            self._bot.sendMessage(self._chat_id, "Infos from the items:", reply_markup={"keyboard":self.create_info_reply_markup()})
+            
+        # /lo: show logics
+        elif command == '/lo':
+            tmp_msg = "";
+            tmp_msg+="Logics:\n"
+            for logic in sorted(self._sh.return_logics()):
+                data = []
+                lo = self._sh.return_logic(logic)
```
ohinckel: The call to `self._sh.return_logics()` can be made forward compatible by using `Logics.return_loaded_logics()` instead.

telegram/README.md
ivan73: "name: ... type: ... knx_dpt: ..." are attributes which are defined in the yaml configuration file of the items.
'telegram_info: my_info_name' is a plugin-specific attribute in the item configuration.

telegram/__init__.py
ivan73: I can not find any more information about this method Logics.return_loaded_logics(). Is this part of SH?

telegram/README.md
ohinckel: Yes, but the quotes are usually not necessary.

telegram/__init__.py
ohinckel: Details are in the developer docs: http://smarthomeng.de/developer/lib/logic.html#module-lib.logic

ivan73: from lib.logic import Logics

self.logics = Logics.get_instance()
for logic in sorted(self.logics.return_loaded_logics()):

AttributeError: 'NoneType' object has no attribute 'return_loaded_logics'

I have to do that Download or update the library (lib.logic)

ohinckel: How do you run your SHNG instance? Usually you have to checkout the recent SHNG [code](https://github.com/smarthomeNG/smarthome/tree/develop) (in develop branch) and inside the plugins directory the SHNG [plugins](https://github.com/smarthomeNG/plugins/tree/develop) (develop branch too).

When developing plugin enhancement you usually use your own repository checkout of the plugins repo.

When the method is missing it means that you do not have a recent develop checkout of the core, because the [current version](https://github.com/smarthomeNG/smarthome/blob/develop/lib/logic.py#L587) contains this.

ivan73: ok, I use SH-Master 1.4.1  I could have imagined that.
I first have to install the developer version parallel.

ohinckel: This is also in the master branch of SHNG - but maybe not in your fork (beacuse it was not updated).

If this all is too complicated we can merge it and refactor this later.

ivan73: at which point in the code I place "self.logics = Logics.get_instance ()". If I do this in the "__init__" function, then the result of .get_instance () is "None"?

msinn: Because logics are initialized after plugins and items are loaded and initialized, you can't put it in the `__init__` routine. Place it at the start of the run method.

ivan73: Many Thanks. I now have access to logic via the new lib.logic-api. I hope it fits. At least it works for me. Sorry for my "good" english

ohinckel: No problem, thanks for your adjustments! Just two notes:
* [line 96](https://github.com/smarthomeNG/plugins/pull/108/commits/14b203678eb77f619149857b6044c2dce894cff5#diff-f2dfe37a3fe68465cab83f206c469e45R96) contains a TAB character for indentation, usually we use 4 spaces (instead of one TAB)
* [line 254](https://github.com/smarthomeNG/plugins/pull/108/commits/14b203678eb77f619149857b6044c2dce894cff5#diff-f2dfe37a3fe68465cab83f206c469e45R254) contains `self._sh.return_logic()` which can also be replaced with `self.logics.return_logic()`

ohinckel:
Someone able to test it before we merge?

bmxp:
Right now I can't test it. I think in the nearer future this plugin should use [python-telegram-bot](https://github.com/python-telegram-bot/python-telegram-bot/) instead of telepot to get rid of the connection problems reported by some users and the 3 Threads that will not stop together with a stop of SmartHomeNG.

ohinckel:
AFAIS, the problem is not related to this PR. So I think we can merge it, right?

bmxp:
The merge should be fine. At the moment @ivan73 has presented a [link to his repo](https://github.com/ivan73/plugins/blob/python-telegram-bot-API/telegram/__init__.py) in [knx-user-forum](https://knx-user-forum.de/forum/supportforen/smarthome-py/922830-telegram-alternative-zu-whatsapp) and he implemented the above change from telepot to python-telegram-bot. I am sure he will start a pull request as soon as this is running stable.

PR #110, opened 2017-12-26
Update rcswitch-pi

hasenradball:
# __init__.py
- add import os

# Readme.md
- add troubleshooting chapter

### DIFF
--- a/rcswitch/README.md
+++ b/rcswitch/README.md
@@ -223,7 +223,7 @@ rc_code = code of device [00000 - 11111]
 ## Troubleshooting
 If the switch does not work, but you are sure that the installation was done properly, make sure that the user (normally) smarthome is part of the group gpio.
 
-If not it ist easy to do this.
+If not, it is easy to do this.
 
 ```
 sudo usermod -aG gpio smarthome
--- a/rcswitch/README.md
+++ b/rcswitch/README.md
@@ -220,8 +220,9 @@ rc_code = code of device [00000 - 11111]
 
 ----------------------------
 
-#### Troubleshooting
+## Troubleshooting
 If the switch does not work, but you are sure that the installation was done properly, make sure that the user (normally) smarthome is part of the group gpio.
+
 If not it ist easy to do this.
 
 ```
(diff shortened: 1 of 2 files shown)


ohinckel:
Thanks for this update. I'm also using rc-switch to trigger my powerplugs, but with a small [enhancement](https://github.com/r10r/rcswitch-pi/pull/16) to automatically support different types with `send` command (by specifying the type of powerplug as argument).

But it seems that this library isn't maintained anymore, so I suggest to reference the more up-to-date version: https://github.com/sui77/rc-switch

Most parts are the same but I guess it supports more diffferent types of powerplugs. What do you think? Does it matter which rc-switch library is referenced in the documentation? Did you tried already the other version (or any other more recently updated)?

hasenradball:
Hi ohinckel, thank you for your comment. The plugin was written from my friend and my colleague. He was very busy the last year, getting a baby and married and so on. But actually we thought that there was no potential for updates th last time.
But in found some issue, if you do not set any optional parameter in the plugin.conf a failure raises due to the missing import os.
One thing what is very fine, he did a special solution in that case that you can trigger many plugs at the same time an the signal will not getting lost, so that ervery plug switches.

I would think it is worth to keep this plugin.
I have not seen your solution but I will have a look in the next days.
Maybe we can discuss to bring these solutions togehter.

I will drink a beer in the next days with him and we will have a look at you solution.

Which kind of plugs does your solution support?

Propose from my side merge in the small issue and we will have a look to your solution.
We was not aware that there is an another one.
Best Regards
Frank

ohinckel:
Great news, @hasenradball. First of all: I don't want to generate additional work for somebody. I only want to avoid double work. So I guess it makes sense to have a look at the new project (which is maintained and my have support for other plugs too).

I only added support for REV switches (some really old one) and keep the current supported one. The command invocation for `send` was then something like this: `send [dip|rev] [device] [state]`. But don't give it too much attention.

The point is, we should stick on some implementation which is still in development. If we use the "new" version I could also provide support for REV switches (if it is not already there).

For the moment, I think we can merge your changes. Thanks so far!

PR #114, opened 2018-01-03
Enocean: update of plugin.yaml and readme.md

hasenradball:
# plugin.yaml
- update parameters
- update of attributes

# Readme.md
- rework the readme.md
- replace aschwith and Rbudde as maintainer and tester

### DIFF
--- a/enocean/README.md
+++ b/enocean/README.md
@@ -9,6 +9,7 @@ If you have special hardware not supported yet please feel free to improve and c
 
 ## Version / Change History
 Version: 1.6
+
 Change History: actually not maintained.
 
 ## Requirements
@@ -101,7 +102,7 @@ STATUS = handle_status
 ### items.yaml / items.conf (deprecated)
 
 #### Attributes
-For attributes see the examples.
+For attributes have a look to the examples.
 
 #### Example item.yaml
 ```
@@ -119,7 +120,7 @@ Enocean_Item:
             type: bool
             enocean_rx_key: STATUS
     
-    FT55switch
+    FT55switch:
         enocean_rx_id: 012345AA
         enocean_rx_eep: F6_02_03
             up:
@@ -438,6 +439,7 @@ Enocean_Item:
 ### Add new listening enocean devices
 
 You have to know about the EnOcean RORG of your device (please search the internet or ask the vendor). 
+
 Further the RORG must be declared in the plugin.
 
 The following status EEPs are supported:
@@ -458,7 +460,7 @@ The following status EEPs are supported:
 * F6_02_03		2-Button-Rocker, Status feedback from manual buttons on different actors, e.g. Eltako FT55, FSUD-230, FSVA-230V, FSB61NP-230V or Gira switches.
 * F6_10_00		Mechanical Handle (value: 0(closed), 1(open), 2(tilted)
 ```
-A complete list of available EEPs is documented at [EnOcean Alliance](http://www.enocean-alliance.org/eep/)
+A complete list of available EEP's is documented at [EnOcean Alliance](http://www.enocean-alliance.org/eep/)
 
 
 ### Send commands: Tx EEPs
@@ -477,8 +479,14 @@ The optional ref_level parameter defines default dim value when dimmer is switch
 ### Learning Mode
 
 Devices that shall receive commands from the smarthome plugin, i.e. the encoean gateway must be subscribed first.
-Generally follow the teach in procedure as described by enocean.
-Usually, the enocean device, e.g. enocean actor, is set to teach in mode. See the manual of the respective device for further information. Once being in teach in mode, trigger a learn-in command from smarthomeNG.
+Generally follow the teach in procedure as described by enocean:
+1. set the enocean device/actor into learn mode
+2. send the learn telegram
+3. exit the learn mode of the actor
+
+Usually, the enocean device, e.g. enocean actor, is set to teach in mode.
+See the manual of the respective device for further information.
+Once being in teach in mode, trigger a learn-in command from smarthomeNG.
 
 In order to send a special learning message, start smarthome with the interactive console:
 
--- a/enocean/README.md
+++ b/enocean/README.md
@@ -2,6 +2,7 @@
 
 ## Description
 This plugin allows you to interact with enocean devices via SmarthomeNG and SmartVisu.
+
 This plugin is still under development.
 
 ## Support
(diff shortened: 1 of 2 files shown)


enocean/README.md
```diff
@@ -1,23 +1,63 @@
...
+
+This plugin is still under development.
+
+## Support
+If you have special hardware not supported yet please feel free to improve and contribute!
+
+## Version / Change History
+Version: 1.6
+
+Change History: actually not maintained.
```
ohinckel: Actually vs. currently?

enocean/README.md
```diff
@@ -1,23 +1,63 @@
...
+When controlling multiple devices, it is recommended to use the enocean-adapter's Base-ID (not Unique-ID or Chip-ID) as transmitting ID.
+For further information regarding the difference between Base-ID and Chip-ID, see
+[Knowledge Base](https://www.enocean.com/en/knowledge-base-doku/enoceansystemspecification%3Aissue%3Awhat_is_a_base_id/)
 
-When controlling multiple devices, it is recommended to use the stick's BaseID (not ChipID) as transmitting ID.
-For further information regarding the difference between BaseID and ChipID, see [Knowledge Base](https://www.enocean.com/en/knowledge-base-doku/enoceansystemspecification%3Aissue%3Awhat_is_a_base_id/)
+With the specification of the Base-ID, 128 different transmit ID's are available, ranging between Base-ID and Base-ID + 127.
 
-With the specification of the BaseID, 128 different transmit IDs are available, ranging between BaseID and BaseID + 127.
+##### How-To Get the ID of an EnOcean device
```
ohinckel: What is the correct notation for enocean? EnOcean, ENOCEAN, enocean, ...? We should stick to one notation and use it everywhere in the documentation. Only a minor thing, but makes the documentation looking cleaner.

enocean/README.md
```diff
@@ -26,208 +66,46 @@ With the specification of the BaseID, 128 different transmit IDs are available,
...
-right rocker down = BI
-right rocker up = BO
+AI = left rocker down
+A0 = left rocker up
+BI = right rocker down
+B0 = right rocker up
 ```
 
-The following example is for a rocker/switch with two rocker and 6 available combinations (EEP F6_02_03).
+The following example example explaines the button shortcut and its meaning for a rocker/switch with two rocker and 6 available combinations (EEP F6_02_03).
```
ohinckel: Double "example" in this line.

enocean/README.md
```diff
@@ -396,9 +274,173 @@ Enocean_Item:
...
+            enocean_tx_id_offset = 2
+    
+    [[actorD2]]
+        enocean_rx_id = FFDB7381
+        enocean_rx_eep = D2_01_07
+        [[[move]]]
+            type=bool
+            enocean_rx_key = STAT
+            enocean_tx_eep = D2_01_07
+            enocean_pulsewidth = 0.1  // optional; turn off after 0.1 sed
```
ohinckel: We could use there `#` instead of `//` to indicate a comment.

enocean/README.md
```diff
@@ -419,7 +461,7 @@ The following status EEPs are supported:
 * F6_02_03		2-Button-Rocker, Status feedback from manual buttons on different actors, e.g. Eltako FT55, FSUD-230, FSVA-230V, FSB61NP-230V or Gira switches.
 * F6_10_00		Mechanical Handle (value: 0(closed), 1(open), 2(tilted)
 ```
-A complete list of available EEPs is documented at [EnOcean Alliance](http://www.enocean-alliance.org/eep/)
+A complete list of available EEP's is documented at [EnOcean Alliance](http://www.enocean-alliance.org/eep/)
```
ohinckel: I guess in this case "EEPs" is correct, "EEP's" not.

enocean/plugin.yaml
```diff
@@ -2,23 +2,93 @@
...
+        type: str
+        default: /dev/ttyAMA0
+        description:
+            de: 'Name der Schnittstelle an der sich der Enocean Adapter befindet'
+            en: 'name of the port where the enocean adapter is plugged'
+        valid_list:
+          - /dev/ttyAMA0
+          - /dev/ttyS0
+          - /dev/ttyS1
+          - ...
```
ohinckel: AFAIK, the `valid_list` contains possible values for this parameter. In this case you will only be allowed to specify the three mentioned devices and `...`.

And I think listing devices there is not a good idea since the names of the devices could be very different (depending on the system and the hardware you're using).

Maybe we could introduce a pattern support for `valid_list` setting which can be used to specify a regex pattern which will be matched against the values. But this is something @msinn can decide if this is a good or bad idea.

enocean/plugin.yaml
```diff
@@ -2,23 +2,93 @@
...
+        mandatory: 'True'
+          
+    tx_id:
+        type: int
+        default: FFFF4680
+        description:
+            de: 'Unique-ID oder Base-ID des enocean Moduls (8-stelliger hex Wert)'
+            en: 'Unique-ID or Base-ID of the enocean Module(8-digit hex value)'
```
ohinckel: Blank missing here :wink:

enocean/plugin.yaml
```diff
@@ -2,23 +2,93 @@
...
+            de: 'ID des enocean Gerätes (Aktor oder Sensor)'
+            en: 'ID of the enocean device (actor or sensor)'
+        mandatory: 'True'
+        
+    enocean_rx_eep:
+        type: str
+        default: F6_02_03
+        description:
+            de: 'EEP Profile des Response Telegramms'
+            en: 'EEP Profile of the repeat meassage'
```
ohinckel: Repeat vs. response?

enocean/plugin.yaml
```diff
@@ -2,23 +2,93 @@
...
+            en: 'actor button name of the enocean device'
+        valid_list:
+          - A
+          - AI
+          - A0
+          - B
+          - BI
+          - B0
+          - ...
+        mandatory: 'False'
```
ohinckel: You don't need to explicitly quote this boolean values.

hasenradball:
Hello ohinckel,
thank your for your review!
Your findings are correct, files were reworked.

hasenradball:
Hello ohinckel,

I want to discuss again with you about the rc-switch topic.
Were can we do this?
Regrads
Frank

enocean/plugin.yaml
ohinckel: The quotes are still there.

ohinckel:
I suggest to use the gitter chat.

hasenradball:
Does it look good now for your Point of view?

PR #119, opened 2018-01-14
Harmony plugin bugfixes

pfischi:
Bugfix: 'unknown activity' in debug log
updated documentation

### DIFF
--- a/harmony/__init__.py
+++ b/harmony/__init__.py
@@ -14,7 +14,7 @@ class Harmony(SmartPlugin):
     PLUGIN_VERSION = "1.4.0"
     ALLOW_MULTIINSTANCE = False
 
-    def __init__(self, sh, harmony_ip, harmony_port=5222, sleekxmpp_debug=False):
+    def __init__(self, sh):
         self._is_active = False
         self._get_config_reconnect = 300  # scheduler time to get config from Hub and for possible reconnect
         self._scheduler = sched.scheduler(time.time, time.sleep)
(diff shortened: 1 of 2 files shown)


harmony/plugin.yaml
```diff
@@ -2,24 +2,39 @@
...
+    type: int
+    description:
+      de: "(optional) Harmony Hub Port. Default: 5222"
+      en: "(optional) Harmony hub port. Default: 5222"
+
+  sleekxmpp_debug:
+    type: bool
+    description:
+      de: "(optional) Debug-Ausgaben der sleekxmpp-Klasse. Default: false"
+      en: "(optional) Debug print of ht esleekxmpp class. Default: false"
```
ohinckel: The "print of ht eslee" should be "print of the slee".

harmony/__init__.py
```diff
@@ -114,9 +119,12 @@ def _set_current_activity(self, activity=None):
...
 
     def _send_activity(self, activity):
+        activity = int(activity)
         label = "unknown"
-        if activity in self._activities:
-            label = self._activities[activity]
+        for activity_id, activity_name in self._activities.items():
+            if activity_id == activity:
+                label = activity_name
+                break
```
ohinckel: What is the different between the new and the old implementation? At least the add `int` conversion makes sense since the keys are [added as int values](https://github.com/smarthomeNG/plugins/pull/119/files#diff-90b7a0a8b77e4bc5309c66a23782fe53R97).

But shouldn't it be enough to just convert the value to `int` and keep the old implementation in this method?

pfischi: You're right. This is just a relict of a larger code block while debugging. I've changed this to the previous version (+ integer check).

harmony/plugin.yaml
pfischi: done

harmony/plugin.yaml
```diff
@@ -2,24 +2,39 @@
...
+      de: "(optional) Harmony Hub Port. Default: 5222"
+      en: "(optional) Harmony hub port. Default: 5222"
+
+  sleekxmpp_debug:
+    type: bool
+    description:
+      de: "(optional) Debug-Ausgaben der sleekxmpp-Klasse. Default: false"
+      en: "(optional) Debug print of the sleekxmpp class. Default: false"
+
 item_attributes:
```
ohinckel: Just noticed: Do you know about the `default` setting of [parameter configuration](http://smarthomeng.de/developer/development_plugin/plugin_metadata.html)? This way you don't need to handle this in your plugin code.

pfischi: Thanks for your hint. I have implemented the new metadata handling.

ohinckel:
Thanks for updating this PR. I think we can merge this now.

PR #120, opened 2018-01-29
JVC D-ILA Control Plugin

Civer666:
Hey guys,
i want to add my [JVC D-ILA Control Plugin](https://knx-user-forum.de/forum/supportforen/smarthome-py/1188479-plugin-steuerung-von-jvc-d-ila-projektoren) to smarthomeNG plugins!
Is there a way to get write access, too?

Greetings, KNXinsHaus

### DIFF
--- a/jvcproj/README.md
+++ b/jvcproj/README.md
@@ -4,7 +4,7 @@ With this plugin you can control JVC D-ILA projectors over TCP by using the "JVC
 
 ## Notes
 I use the plugin to control my JVC DLA- X5000 and transfer raw gammatables generated with jvcprojectortools by arvehj. Many thanks to arvehjs groundwork so i could have a look in the communication procedure!
-You can use all operation commands given in the "JVC External Control Command Communication Specification". Wrong or unknown commands will be ignored by the plugin and the projector. Most of the commands won`t be acknowledged when the projector has no input signal. Request commands are not yet supported.
+You can use all operation commands given in the "JVC External Control Command Communication Specification". Wrong or unknown commands will be ignored by the plugin and the projector. Most of the commands will not be acknowledged when the projector has no input signal. Request commands are not yet supported.
 The gamma table configuration files can be raw gamma table files (loaded from the projector with jvcprojectortools) or the configuration files that are generated with jvcprojectortools when you save your parameters. When using these files only the deposited raw gamma data ("table"- data) is needed and will be transfered.
 
 The plugin should work from the projector generation X3/X7/X9 up to X5900/X7900/X9900 and surely for upcoming generations, too.
@@ -19,13 +19,13 @@ Please use [this thread for support, questions, feedback etc.](https://knx-user-
 jvcproj:
     class_name: JVC_DILA_Control
     class_path: plugins.jvcproj
-    host: 1.1.1.1 # host adress of the projector
+    host: 1.1.1.1 # host address of the projector
     port: ... # optional, default port is 20554
     gammaconf_dir: ... # optional, location gamma table configuration files
 ```
 
 #### Attributes
-  * `host`: the host adress of the projector
+  * `host`: the host address of the projector
   * `port`: the port the projector listens to. Default is 20554
   * `gammaconf_dir`: location where the gammatable configuration files are saved. Default is '/usr/local/smarthome/etc/jvcproj/'
 
@@ -51,21 +51,21 @@ jvcproj:
         type: bool
         visu_acl: rw
         jvcproj_cmd: 21 89 01 00 00 0A
-        enforce_updates: 'True'
+        enforce_updates: True
 
     pwon:
     # command Power ON
         type: bool
         visu_acl: rw
         jvcproj_cmd: 21 89 01 50 57 31 0A
-        enforce_updates: 'True'
+        enforce_updates: True
 
     pwoff:
     # command Power OFF
         type: bool
         visu_acl: rw
         jvcproj_cmd: 21 89 01 50 57 30 0A
-        enforce_updates: 'True'
+        enforce_updates: True
 
     picmode:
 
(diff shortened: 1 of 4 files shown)


jvcproj/README.md
```diff
@@ -0,0 +1,251 @@
+# JVC D-ILA Control
+
+With this plugin you can control JVC D-ILA projectors over TCP by using the "JVC External Control Command Communication Specification" and transfer gammatables generated with jvcprojectortools.
+
+## Notes
+I use the plugin to control my JVC DLA- X5000 and transfer raw gammatables generated with jvcprojectortools by arvehj. Many thanks to arvehjs groundwork so i could have a look in the communication procedure!
+You can use all operation commands given in the "JVC External Control Command Communication Specification". Wrong or unknown commands will be ignored by the plugin and the projector. Most of the commands won`t be acknowledged when the projector has no input signal. Request commands are not yet supported.
```
ohinckel: You should use simple quotes, not back ticks (could lead to problems when rendering this page).

jvcproj/README.md
```diff
@@ -0,0 +1,251 @@
...
+jvcproj:
+    class_name: JVC_DILA_Control
+    class_path: plugins.jvcproj
+    host: 1.1.1.1 # host adress of the projector
+    port: ... # optional, default port is 20554
+    gammaconf_dir: ... # optional, location gamma table configuration files
+```
+
+#### Attributes
+  * `host`: the host adress of the projector
```
ohinckel: Typo in "address", two times (see above).

jvcproj/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+```yaml
+jvcproj:
+
+    testcmd:
+    # command NULL
+        type: bool
+        visu_acl: rw
+        jvcproj_cmd: 21 89 01 00 00 0A
+        enforce_updates: 'True'
```
ohinckel: Do you really need these single quotes around "True"?

jvcproj/__init__.py
```diff
@@ -0,0 +1,367 @@
...
+        Write items values
+        
+        :param item: item to be updated towards the plugin
+        :param caller: if given it represents the callers name
+        :param source: if given it represents the source
+        :param dest: if given it represents the dest
+        """
+        if item():
+            if self.has_iattr(item.conf, 'jvcproj_cmd'):
+                if self.get_iattr_value(item.conf, 'jvcproj_cmd') == 'None':
```
ohinckel: Why do you need a special 'None'  check? Is the `has_iattr()` check one line above not enough? Do you really want to set the attribute to value `"None"`?

jvcproj/plugin.yaml
```diff
@@ -0,0 +1,25 @@
+# Metadata for the Smart-Plugin
+plugin:
+    # Global plugin attributes
+    type: interface                 # plugin type (gateway, interface, protocol, system, web)
+    description:
+        de: 'Plugin um JVC D-ILA Projektoren zu kontrollieren und durch jvcprojectortools erstellte Gammakurven zu uebertragen'
+        en: 'plugin to control JVC D-ILA projectors and transfer gammatables generated with jvcprojectortools'
+    maintainer: NPortz
+    tester:
```
ohinckel: Do you now any other testers to list here?

jvcproj/plugin.yaml
```diff
@@ -0,0 +1,25 @@
...
+        de: 'Plugin um JVC D-ILA Projektoren zu kontrollieren und durch jvcprojectortools erstellte Gammakurven zu uebertragen'
+        en: 'plugin to control JVC D-ILA projectors and transfer gammatables generated with jvcprojectortools'
+    maintainer: NPortz
+    tester: 
+#    keywords: iot xyz
+#    documentation: https://github.com/smarthomeNG/smarthome/wiki/CLI-Plugin        # url of documentation (wiki) page
+#    support: https://knx-user-forum.de/forum/supportforen/smarthome-py
+
+    version: 1.0.0                 # Plugin version
+    sh_minversion: 1.1             # minimum shNG version to use this plugin
```
ohinckel: I guess SH 1.1 does not have SmartPlugin support. I would increase at least to 1.3.

jvcproj/plugin.yaml
```diff
@@ -0,0 +1,25 @@
...
+#    sh_maxversion:                 # maximum shNG version to use this plugin (leave empty if latest)
+    multi_instance: False          # plugin supports multi instance
+    classname: JVC_DILA_Control    # class containing the plugin
+
+parameters:
+    # Definition of parameters to be configured in etc/plugin.yaml
```
ohinckel: If you document the plugin parameters here you could benefit from some default values also declare here instead of checking in the code. Take a look into the [docs](http://smarthomeng.de/developer/development_plugin/plugin_metadata.html) to see how it works.

jvcproj/plugin.yaml
```diff
@@ -0,0 +1,25 @@
...
+    
+item_attributes:
+    # Definition of item attributes defined by this plugin
```
ohinckel: Adding documentation of the item attributes here would also help to improve the documentation.

jvcproj/__init__.py
Civer666: smartplugin.py returns True if item exists. But when no command is declared (jvcproj_cmd = ..) the returned value will be "None" or not? So you can debug a message to have a look in the item because no command is given there. That my intension.

jvcproj/plugin.yaml
Civer666: No, sorry. Of course i know some People with that kind of projector. But most of them use Crestron to control the projector. I`m the only one who tested it at the moment.

jvcproj/plugin.yaml
Civer666: Yes, thats right! The complete plugin.yaml has to be revised anyway.

jvcproj/README.md
Civer666: No, i do not think so! I think these quotes exist in my items.conf since the beginning of smarthome.py! I will change it!

jvcproj/plugin.yaml
Civer666: Yes, i will do that!

Civer666:
Ok. I´ve updated the Readme.md, added parameters to plugin.yaml and removed the option to set a port because it is always fixed to 20554 (since years) by JVC.
Only one thing: i wont set item_attributes to the plugin.yaml because the attributes are raw hex commands in variable lenght. It depends on beamer generation, too.

jvcproj/plugin.yaml
```diff
@@ -19,7 +19,20 @@ plugin:
...
-    
+    host:
+        type: ip
+        default: 0.0.0.0
+        description:
+            de: 'Die IP Adresse des JVC Projektors ist anzugeben.'
+            en: 'Define IP address of your JVC projector.'
+
+    gammaconf_dir:
+        type: dict
```
ohinckel: This should be type `str` because you configured a path, not a dictionary directly.

ohinckel:
Adding configuration it `item_attributes` is (currently) only used for documentation purpose. So I suggest to add them too. You don't need to provide a default value. Its enough to describe and declare the type of it in the `plugin.yaml`.

Civer666:
Maybe a stupid question but i have to ask it! ;) Are "jvcproj_cmd" and "jvcproj_gamma" boolean or a string items? The item itself is boolean but the declared commands are strings! Normally i would say they're boolean...

ohinckel:
When you talk about "items" you mean smarthomeNG item or just item attributes? In the first case the `type` defines the type of the item's value and not the type of any item attribute.

```yaml
  items:
    cmd:
      type : bool
      jvcproj_cmd: XY XY XY
    gamma:
      type: bool
      jvcproj_gamma: ....
```

When you are talking abou the `item_attributes` section of `plugin.yaml` then you need to define the type of the attributes. In the case above both seems to be a `str`.

Civer666:
Ok, in the end i did it right yesterday. I will push the changes this evening.

ohinckel:
I guess we can merge this now. Opinions?

Civer666:
From my side: yes you can. The plugin works fine. I just haven`t found any other testers yet.

PR #125, opened 2018-02-15
python-telegram-bot api

ivan73:
changed to use whith python-telegram-bot API
add ITEM_ATTR_MATCHREGEX - check a value against a condition before sending the message
add ITEM_ATTR_TEXT - write message-text into the SH-item whit this attribut


ohinckel:
@ivan73, just let you know: your enhancement was implemented (manually without PR merge, see mentioned commit). So next time just be more patient ;-)

PR #129, opened 2018-03-05
Sonos update / bugfix

pfischi:


sonos/__init__.py
```diff
@@ -343,7 +343,11 @@ def dispose(self):
             return
 
         for subscription in self._events:
-            subscription.unsubscribe()
+            try:
+                subscription.unsubscribe()
+            except:
+                continue
+
```
ohinckel: In which case an exception is thrown by `unsubscribe()` method? Is it OK to simply ignore the exception or should it be logged so users can find problems with it?

pfischi:
This is just a quick and dirty hack to prevent a ugly failure. I'am currently working on a slightly better way to handle the event subscription / unsubscription.

pfischi:
Some git struggles here. I'll do a new pull request.

PR #130, opened 2018-03-07
Sonos bugfix

pfischi:


pfischi:
I had some trouble with my git repo, so some commits are duplicates. So just ignore this.

PR #131, opened 2018-03-11
HUE-plugin: fix readme (lists are always required) and authorize-func

robert-budde:
Fix two severe flaws which are no good entry for novice users without debug-experience:
- Readme mentions wrong configuration for easy setups - lists are always required, checks are notr obvious or not working
- authorize-func crashes because of unproper conversion from bytes to string, after it fails at it does not use format mandated by https://www.developers.meethue.com/documentation/getting-started

### DIFF
--- a/hue/README.md
+++ b/hue/README.md
@@ -60,8 +60,10 @@ Minimal configuration for single bridge an default settings
 HUE:
     class_name: HUE
     class_path: plugins.hue
-    hue_user: 38f625a739562a8bd261ab9c7f5e62c8
-    hue_ip: 192.168.2.2
+    hue_user:
+      - 38f625a739562a8bd261ab9c7f5e62c8
+    hue_ip:
+      - 192.168.2.2
 ```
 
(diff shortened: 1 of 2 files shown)


robert-budde:
this also applies to develop, yet it should be fixed in master for sure...

msinn:
@ohinckel Kannst Du mal schauen was mit den travis tests schief gelaufen ist? Die Meldungen schreiben was von Ruby????

ohinckel:
Die Aenderungen sind im `master` Branch gemacht und basieren auf einer alten Version (dort fehlt die Travis-Konfiguration und es wird ein falscher Build gestartet).

@robert-budde, kannst du bitte die aktuellen Aenderungen dir holen und ein rebase auf deinen Branch machen?

msinn:
Parameter conversion for lists: parameters of type str are converted to lists (of len 1) is handled globally in lib.metadata.

Fixing the authorize function was done by a separate commit.

PR #160, opened 2018-08-31
GPIO Plugin: Added webinterface and optimized code and logging

onkelandy:


gpio/plugin.yaml
```diff
@@ -11,15 +11,21 @@ plugin:
...
     # Definition of parameters to be configured in etc/plugin.yaml
-    
+    mode:
+        type: str
+        default: 'BOARD'
+        mandatory: True
+        description:
+            de: 'BOARD oder BCM, abhängig davon, nach welcher Methode die Pins zugewiesen werden sollen.'
+            en: 'BOARD or BCM, depending on how to assign the pins.'
+
```
ohinckel: AFAIK, you can specify the list of possible values by using `valid_list` setting.

gpio/__init__.py
```diff
@@ -20,75 +20,98 @@
...
-        self.logger.debug("GPIO: Mode set to {0}".format(self._mode))
-        self.alive = False
-        self._lock = threading.Lock() 
-        
+        self.init_webinterface()
+        self._name = self.get_fullname()
+        if not REQUIRED_PACKAGE_IMPORTED:
+            self.logger.error("{}: Unable to import Python package 'GPIO'".format(self._name))
+            self._init_complete = False
+            return
```
ohinckel: I don't know if this error handling will help people or not. Sure, this prevents a fatal error, but also does not handle such an incomplete plugin setup in other places of this plugin.

Someone is still able to call the `send()` method directly ending with a fatal error. Further an incomplete setup will not avoid that the `parse_item ()` will be called for all the item - which is senseless dude like since the GPIO can not be used.

onkelandy: It's taken from the sample plugin...

gpio/plugin.yaml
onkelandy: Indeed that should be included.. I guess it's case-sensitive? Just commited it.

gpio/__init__.py
onkelandy: Isn't it the same as described here? https://www.smarthomeng.de/fehlerbehandlung-fuer-plugins-bei-nicht-installierten-python-packages

ohinckel:
@onkelandy, can you resolve the conflicts?

onkelandy:
Konflikt behoben.

ohinckel:
Added to the feature list:
https://github.com/smarthomeNG/smarthome/wiki/SmartHomeNG-v1.6---geplante-Features/_compare/45a21d59d9493ac072a5987fb770ce90533a2745...9f821d75ef840316d5661a2d5c4b3a981e47b3eb

PR #163, opened 2018-09-22
drexelundweiss: Updated logging and other stuff based on sample plugin

onkelandy:
- logging name
- init of serial package
- metadata support from plugin.yaml
- readonly items are now queried after update

### DIFF
--- a/drexelundweiss/__init__.py
+++ b/drexelundweiss/__init__.py
@@ -26,6 +26,7 @@ import os.path
 import time
 import string
 import re
+import codecs
 from lib.model.smartplugin import SmartPlugin
 from bin.smarthome import VERSION
 
@@ -44,7 +45,7 @@ class DuW(SmartPlugin):
         if '.'.join(VERSION.split('.', 2)[:2]) <= '1.5':
             self.logger = logging.getLogger(__name__)
         if not REQUIRED_PACKAGE_IMPORTED:
-            self.logger.error("{}: Unable to import Python package 'serial'".format(self._name))
+            self.logger.error("Unable to import Python package 'serial'")
             self._init_complete = False
             return
         try:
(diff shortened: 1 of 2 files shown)


drexelundweiss/__init__.py
```diff
@@ -18,74 +18,90 @@
...
 import string
 import re
 from lib.model.smartplugin import SmartPlugin
+from bin.smarthome import VERSION
+
+try:
+    import serial
+    REQUIRED_PACKAGE_IMPORTED = True
+except Exception:
+    REQUIRED_PACKAGE_IMPORTED = False
```
ohinckel: This could also be solved using a more general implementation.

I don't like general stuff implemented in several plugins.

Maybe the import can be done by a `Plugins. import ()` which returns the imported module or None in case of an Error. It could also log a general message.

drexelundweiss/__init__.py
```diff
@@ -18,74 +18,90 @@
...
+        except Exception as err:
+            self.logger.error("{}: Error on init {}.".format(self._name, err))
+            self._init_complete = False
+            return
         try:
-            self._port = serial.Serial(tty, 115200, timeout=5)
+            self._port = serial.Serial(self._tty, 115200, timeout=5)
         except:
-            self.logger.error("DuW: could not open {}.".format(tty))
+            self.logger.error("{}: could not open {}.".format(self._name, self._tty))
```
ohinckel: Isn't the plugin logger already including the plugin name in the message?

onkelandy: This part was taken from here: https://github.com/smarthomeNG/smarthome/commit/31fb104d03c6d962d3faaef9f68e8bdac96bd3e6

But it looks like this changed meanwhile.. So starting the log entry with the plugin name isn't useful anymore?

onkelandy: Changed logging. Hope it is fine this way.
By the way: logutils verhindert Smarthome Start, weil in Zeile 41 das : hinter else fehlt!

onkelandy:
Yes. should be ready.

ohinckel:
@onkelandy, can you tell me what we can add to the [feature page](https://github.com/smarthomeNG/smarthome/wiki/SmartHomeNG-v1.6---geplante-Features)?

onkelandy:
Beim Updaten eines Readonly-Items wird dessen Wert abgefragt

Mehr ist's nicht, der Rest war nur under the hood.

ohinckel:
OK, danke. Ich habe das Mal aufgenommen:
https://github.com/smarthomeNG/smarthome/wiki/SmartHomeNG-v1.6---geplante-Features/_compare/9f821d75ef840316d5661a2d5c4b3a981e47b3eb...73cbf16b5849ea59216210081bd7e2cc330a7959

PR #172, opened 2018-10-10
Create requirements.txt

henfri:


ohinckel:
The plugin itself does not require any additional module. So there is no need for a requirements file.

It all depends on the user and the database system used by the user. There's no strict dependency on MySQL. One could use another database system too (e.g. sqlite, postgres, ...).

I think a requirements doesn't help very much. At least we want to show users how to use it with a MySQL database to get started.

henfri:
Hello,

I understand -and disagree.
I prefer to install a python module that is not needed over missing a module. In particular:
https://knx-user-forum.de/forum/supportforen/smarthome-py/1275678-docker-container-database-plugin-maria-db

I use the requirements-all.txt when building the Docker-Container. I cannot go through all Plugins and check what modules may be used by the plugin.
And it is not easy for the user to install further modules into the docker container.

onkelandy:
This issue is relevant for quite some plugins (avdevice, uzsu, etc.)
I somehow agree that it's easier and less of a problem to not use a module than to miss a module in case someone wants to use a feature. But then again, the requirements.txt needs to be put in all plugins with optional modules.

henfri:
Hello,

let's make an experiment:
smarthome-ng does need none of the modules that are in the plugin-requirements.txt.

If we follow the argument, then we should not put anything in the requirements.txt, as smarthome-ng can also be used without plugins.

In other words:
I am sure I am installing lots of modules that I do not need already due to the fact that I do not use all plugins.

Greetings,
Hendrik

ohinckel:
> If we follow the argument, then we should not put anything in the requirements.txt, as smarthome-ng can also be used without plugins.

Yes, you're right. But I guess this is just a infrastructure problem. We do not have a dependency system currently which tells the user (or install) the missing dependencies.

I think for the moment it would be the best to put the dependencies into the requirements file for the database plugin. This all only to avoid "getting started problems" of users.

Still they don't be forced to use a MySQL database. If they decide to use PostgreSQL (or other database system) they still have trouble setting this up because the dependencies for them are not listed in the requirements file. And I though the [plugin description](https://github.com/smarthomeNG/plugins/blob/develop/database/README.md) tells the user about which dependency should be installed for which database system.

msinn:
By the way: The requirements that are checked and might prevent SmartHomeNG from starting (if missing), are only requirements by the core and the core modules. 

Plugn requirements are not checked at start time. They are indirectly checked when the plugin is initializeng and the import fails. That exeption is caught and a message with the name of the missing package is logged.

ohinckel:
I don't know if this would be possible but it seems to be a good solution: the modules for SmartHomeNG core could be listed in the `requirements.txt` file. No other modules are listed there anymore.

The modules used by plugins are only listed in their `requirements.txt` file which is checked for missing dependencies by SmartHomeNG when it loads a plugin. As mentioned by @msinn such a check is already implemented (which logs errors to the log file), but it would be great if we also see such an initialization error in the backend frontend where all the plugins are listed. So users can see the initialization error. Further it could also list the missing dependencies for the plugins.

For this PR: instead of just putting the module in the plugin's requirement file we could create requirement files for features of plugins. E.g. on for MySQL (`requirements-mysql.txt`) one for PostgreSQL (`requirements-postgresql.txt`).

onkelandy:
The requirements should be linkeable to an plugin attribute in my opinion. So let's say the attribute "driver" has the value "pymysql" the relevant requirements.txt is read and possible problems are logged. What do you think? I guess there are some plugins that have optional features with then optional requirements..

msinn:
@ohinckel 
> I don't know if this would be possible but it seems to be a good solution: the modules for SmartHomeNG core could be listed in the requirements.txt file. No other modules are listed there anymore.

That's what the new requirements handling is about. The build_requirements.py tool and the test routine run at startup of shng build the files **base.txt** and **all.txt**.

The test routine checks afterwards if the base requirements are met ans exits shng otherwise.

base.txt is created from:
- lib/requirements.txt
- modules/*/requirements.txt

all.txt is created from:
- lib/requirements.txt
- modules/*/requirements.txt
- plugins/*/requirements.txt

At the moment I am thinking about an additional file containing the requirements from base.txt and of the plugins that are configured in etc/plugin.yaml. That file can be used for a complete check on startup of shng and to install required packages without installing packages for all (over 100) plugins.

msinn:
Since pymysql is not a real requirement for the plugin, I close the PR.

henfri:
Sorry, I disagree.

> I prefer to install a python module that is not needed over missing a module. In particular:
> https://knx-user-forum.de/forum/supportforen/smarthome-py/1275678-docker-container-database-plugin-maria-db
> 
> I use the requirements-all.txt when building the Docker-Container. I cannot go through all Plugins and check what modules may be used by the plugin.
> And it is not easy for the user to install further modules into the docker container.

The decision not to include all possible requirements to a plugin leads renders the docker containers almost useless.

I understand that creating a sophisticated requirements-management is out of scope. But I strongly suggest putting all requirements into the requirements.txt.

I think that there are few plugins that have optional requirements. Thus my suggestion will not harm.

Greetings,
Hendrik

msinn:
You can install pymysql if YOU want and need it. I don‘t need pymysql and I don‘t want to be forced to insgall it!

onkelandy:
Es geht ja nicht nur um ihn. Er stellt das docker Image allen bereit. Und da verstehe ich seinen Ansatz schon. 

Wärs nicht möglich und sinnvoll mandatory und optional Pakete im plugin.yaml auszulesen und die ins requirements/xx.txt zu speichern?

msinn:
@onkelandy Ich sehe das ein wenig anders. Bedenke noch mal die folgenden Punkte:

1. Wenn er in der Dockerumgebung MySql nutzt, wird er auch in der Lage sein für den Docker Container pymysql zu installieren. Ich finde es stark übertrieben zu behaupten, wenn wir dieses Requirement (welches aus shNG Sicht kein wirkliches ist) nicht zu implementieren, wären die Docker Container nutzlos. In früheren Versionen von shNG gab es keine requirements.txt. Was shNG dehalb in Docker Umgebungen nicht einsetzbar?

2. Wie hoch ist der Prozentsatz derjenigen, de MySql nutzen im Verhältnis zu den Nutzern der Standard Möglichkeit zur Datenspeicherung (SQLite)?

3. Wie hoch ist der Prozentsatz der Docker Nutzer, die auch noch MySql nutzen?

4. Wann kommt der nächste und fordert, dass wir für das database Plugin das Python Package für PostgreSQL zum Requirement erheben.

5. Welche anderen Plugins kommen in Zukunft mit optionalen Anforderungen, die wie dieser Fall als Spezialbehandlung zu implementieren wären?

henfri:
Hallo,

ich muss sagen, dass ich mich über diese Diskussion etwas ärgere.
Ich maintaine den Docker-Container und kenne mich damit aus. Ich habe ein 'real live'  Problem gezeigt. Es ist kein theoretisches Problem.

Mit dem Argument (3) aus deiner Aufzählung können wir sicherlich 30-50 python-module hinterfragen, denn: Wer nutzt schon alle Plugins. Dennoch installieren wir -im Docker, wie auch in der Komplettanleitung- alle Module.

Daher mein Vorschlag: Wenn ein Plugin ein modul brauchen kann (abhängig von seiner Konfiguration), dann kommt es in requirements.txt.

Beispiel:
In "Word" kommt ja auch keine Meldung dass du erstmal an die Kommandozeile gehen musst und 
"myword install --update "rechtschreibprüfung"" 
eingeben sollst, wenn du die Deutsche Rechtschreibprüfung nutzen willst.
Mein Gott.... Wie groß ist das betreffende modul?

Mir kann es egal sein. Ich habe ein Modul schnell nachinstalliert. Wenn wir sh.ng aber nutzerfreundlich machen wollen, dann so nicht.

Gruß,
Hendrik

ohinckel:
Die einzige Loesung dir mir hierzu einfaellt waere:
1. wir definieren Core-Module die immer installiert werden muessen (`base.txt`)
2. wir definieren Module die fuer Plugins installiert werden muessen (`requirements.txt` im Plugin)
3. wir definieren Module die fuer Funktionen eines Plugins installiert werden muessen (`feature.txt` im Plugin)

Je nach dem was man innerhalb des Plugins aktiviert/nutzt muessen Pakete installiert werden. Evtl. kann man das Backend so erweitern, dass Module bei Aktivierung nachinstalliert werden. Erst danach ist das Plugin nutzbar. Zusaetzlich gibt es das auch fuer Funktionen/Features des Plugins die man im Backend aktivieren kann und ggf. eine Nachinstallation von weitere Modulen durchfuehrt.

Vielleicht laesst sich da auch ein generischer Ansatz innerhalb von SmartHomeNG implementieren:
* man kann ueber eine SHNG API ein Feature mit entsprechenden `requirements.txt` definieren (z.B. fuer den Core koennte man `shng.register_feature('core', ['base.txt'])` aufrufen um das zu registrieren)
* jedes Plugin kann diese Registerung ebenfalls fuer sich nutzen (z.B. `shng.register_feature('plugin.database.mysql' , ['mysql.txt'])`)
* es gibt eine SHNG API die Abhaengigkeiten fuer Features installieren kann (z.B. `shng.setup_features(['core', 'plugin.database.mysql`, ...)`)
* das Backend kann fuer Plugins die Feature registrieren die Aktivierung/Deaktivierung anzeigen
* ggf. kann man in der `plugin.yaml` die Konfiguration hinterlegen
* ggf. kann man in der `plugin.yaml` hinterlegen, welche Einstellungen fuer welches Feature notwendig ist, damit man das im Backend angezeigt werden kann

Das sind nur Ideen, die mir einfallen wie man sowas generisch loesen kann. Das ganze ist allerdings nicht so einfach umsetzbar und vermutlich auch ein groesseres Ding. Andere Loesungen sind meiner Meinung nach aber nur Notloesungen (z.B. Module separat installieren (wer kann das schon ohne groesse Hilfe), alle Module einfach installieren (wozu, wenn ich 90% nicht benoetige)).

PR #181, opened 2018-10-26
Neues Plugin: roomba_980

zapfen83:
Plugin um einen iRobot Roomba Staubsauger der Serie 900 zu steuern

implementiert sind start / stop / dock und ein paar status meldungen

weiteres folgt später


roomba_980/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+from plugins.roomba_980.roomba import Roomba
+from lib.model.smartplugin import SmartPlugin
+from lib.item import Items
+
+class ROOMBA_980(SmartPlugin):
+
+    ALLOW_MULTIINSTANCE = False
+    PLUGIN_VERSION = "1.0.0"
+
+    logger = logging.getLogger(__name__)
```
ohinckel: Logging is [already initialized](https://github.com/smarthomeNG/smarthome/blob/develop/lib/model/smartplugin.py#L59) by `SmartPlugin`.

roomba_980/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def parse_item(self, item):
+        if self.has_iattr(item.conf, 'roomba_980'):
+            item_type = self.get_iattr_value(item.conf, 'roomba_980')
+            if item_type == "start" or item_type == "stop" or item_type == "dock":
+                return self.update_item
+
+            self._status_items[item_type] = item
+
+            self.logger.debug('{} item gefunden {}'.format(item_type, item))
```
ohinckel: Usually we use english when writing messages to logs.

roomba_980/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def run(self):
+        self.scheduler_add(__name__, self.get_status, prio=5, cycle=self._cycle, offset=2)
+        self.alive = True
```
ohinckel: I would suggest to not use `__name__` as plugin name. Instead use a name which identifies the task (e.g. "read_status"). The plugin name is added automatically by the base class.

roomba_980/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+    def stop(self):
+        self.scheduler.remove(__name__)
+        self.myroomba.disconnect()
+        self.alive = False
+
+    def __call__(self):
+        pass
+
+    def update_item(self, item, caller=None, source=None, dest=None):
+        if caller != __name__ and self.has_iattr(item.conf, 'roomba_980'):
```
ohinckel: The `has_iattr()` check is not required here because the `update_item()` method is only called for items which are handled in the `parse_item()` method. This method already checks for the item attribute.

Or is there any other situation where this method get called without having this item configured?

roomba_980/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+            self.logger.debug('item_update {} '.format(item))
+            if self.get_iattr_value(item.conf, 'roomba_980') == "start":
+               if item() == True:
+                   self.send_command("start")
+            if self.get_iattr_value(item.conf, 'roomba_980') == "stop":
+               if item() == True:
+                   self.send_command("stop")
+            if self.get_iattr_value(item.conf, 'roomba_980') == "dock":
+               if item() == True:
+                   self.send_command("dock")
```
ohinckel: This list of `if` statements can use `elif` for small performance improvement.

roomba_980/plugin.yaml
```diff
@@ -0,0 +1,96 @@
...
+    description:
+      de: "Stoppt die Reinigung"
+      en: ""
+
+  dock:
+    type: bool
+    description:
+      de: "Roomba fährt zur Homebase"
+      en: ""
+
```
ohinckel: Are you able to add the english descriptions?

ohinckel:
Great enhancement, thanks for your work! Maybe you're able to fix some typos in the README file and check my comments?

roomba_980/__init__.py
ohinckel: Thanks for fixing this in the meantime.

roomba_980/plugin.yaml
ohinckel: Thanks for extending the documentation.

roomba_980/__init__.py
ohinckel: New code you committed looks better. Thanks!

roomba_980/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+            self._status_items[item_type] = item
+
+            self.logger.debug('found item {} with function {}'.format(item, item_type))
+
+    def run(self):
+        self.scheduler_add('get_status', self.get_status, prio=5, cycle=self._cycle, offset=2)
+        self.alive = True
+
+    def stop(self):
+        self.scheduler.remove(__name__)
```
ohinckel: I think you need to update the name of tell to remove too.

roomba_980/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+
+        for status_item in self._status_items:
+          if status_item == "status_batterie":
+             self._status_items[status_item](status['state']['reported']['batPct'],__name__)
+          if status_item == "status_bin_full":
+             self._status_items[status_item](status['state']['reported']['bin']['full'],__name__)
+          if status_item == "status_cleanMissionStatus_phase":
+             self._status_items[status_item](status['state']['reported']['cleanMissionStatus']['phase'],__name__)
+          if status_item == "status_cleanMissionStatus_error":
+             self._status_items[status_item](status['state']['reported']['cleanMissionStatus']['error'],__name__)
```
ohinckel: If it's not too much work I would to change this to use `elif` here too.

roomba_980/plugin.yaml
```diff
@@ -0,0 +1,96 @@
+# Metadata for the classic-plugin
+plugin:
+    # Global plugin attributes
+    type: interface                 # plugin type (gateway, interface, protocol, system, web)
+    description:
+        de: 'Anbindung von iRobot Roomba Staubsaugern der Serie 900'
+        en: ''
```
ohinckel: Can you add an English plugin description Here?

roomba_980/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+
+class ROOMBA_980(SmartPlugin):
+
+    ALLOW_MULTIINSTANCE = False
+    PLUGIN_VERSION = "1.0.1"
+
+    myroomba = None
+
+    def __init__(self, sh, adress=None, blid=None, roombaPassword=None, cycle=900):
+        self._sh = sh
```
ohinckel: I think this variable to SHNG instance is not used anywhere and can be removed.

psilo909:
will anything happen here in the future?

PR #182, opened 2018-10-28
[Bug] Fix "Rules not evaluating with new SmarthomeNG 1.5 master"

i-am-offline:
Fix the issue described in https://github.com/i-am-offline/smarthome.plugin.autoblind/issues/24

Reason was, that in the item class "_eval_trigger" seems to be replaced by "_trigger".

### DIFF
--- a/stateengine/StateEngineItem.py
+++ b/stateengine/StateEngineItem.py
@@ -254,7 +254,7 @@ class SeItem:
         self.__item._enforce_updates = True
 
         # set "eval" for item if initial
-        if self.__item._eval_trigger and self.__item._eval is None:
+        if self.__item._trigger and self.__item._eval is None:
             self.__item._eval = "1"
 
         # Check scheduler settings and update if requred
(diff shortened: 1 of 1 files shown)


ohinckel:
Because I just noticed it now: this plugin is using a log of "internal" fields from other objects (e.g. `lib.Item`). We should provide access methods in these objects instead of directly accessing private fields.

@msinn, is the `Item._eval_trigger` field used anymore in SmartHomeNG?

msinn:
It got renamed because the triggers are not only for evals, but for a couple of other things. So: No, the new name is _trigger. Since it is a private var, it shouldn‘t be used from outside.

If information about items is needed that is not available now (besides private vara of that class) a feature request should be opened. Based on that request a stabe interface to those informations can be implemented.

i-am-offline:
Feature request SmarthomeNG/smarthome#296 has been created

msinn:
Feature request is ready to test. Have a look at smarthomeNG/smarthome#296 .

PR #188, opened 2018-11-19
cyclic option added 

henfri:
See previous pull request for documentation


onkelandy:
How do we proceed here? Can you try an update as smaiLee mentioned @henfri ?
Anyhow, please delete your two PRs and make a new and clean one with two or one commits.

henfri:
Hello,

unfortunately, I don't  have time in the next four weeks to test enforce_update+cycle.
I think, what you mean is this:
https://www.smarthomeng.de/user/konfiguration/items_standard_attribute_cycle.html?highlight=cycle
Right?

I think that the chante I propose is small enough and it is more intuitive. But I leave it up to you.

Regards,
Hendrik

onkelandy:
I've tested the enforce_updates and cycle alternative - works totally fine. So having an additional option in the plugin makes it more complicated than necessary.

If such an option really makes things better in your opinion feel free to create a new, clean PR.

PR #203, opened 2018-12-29
Database Plugin: avoid error on dump when connection is lost

onkelandy:


onkelandy:
A message in the log and esp. in the web interface would be good indeed. Right now the exception leads to an "endless hang" when dumping from the webif and an 500 error.

ohinckel:
OK, in this case we should check why it hangs. Maybe there is another locking/releaseing issue.

ohinckel:
Are we able to reproduce the error to identify why it hangs? In case of an exception an error should be logged.

onkelandy:
I'm not sure if we can easily recreate the problem. I just know I had it and the simple fix fixed it ;) I've added error messages now.

ohinckel:
Of course solving problems by just ignoring them would "fix" it. But still, the error is just ignored and I think ignoring errors is the wrong solution (like try/catch block around the whole program and not handling the exception at all).

If we can't reproduce the problem, does it still exist?

PR #204, opened 2018-12-30
Plugin zum Auslesen von 1-Wire-Sensoren direkt am RPi angeschlossen

soft-rudi:
Nach mehrfachem Wunsch stelle ich dieses Plugin zur Verfügung. 
Dies ist mein erster pull request und ich weiß nicht ob es der richtige Weg ist ....

LG

### DIFF
--- a/rpi1wire/README.md
+++ b/rpi1wire/README.md
@@ -1,0 +1,181 @@
+# Raspberry Pi 1-Wire plugin
+
+# Requirements
+
+This plugin has been developed for the Raspberry Pi. It allows easy access to 1-Wire - sensors.
+We tested the plugin with the Raspberry Pi B and the temperature sensor DS18B20.
+
+## Supported Hardware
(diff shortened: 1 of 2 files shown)


psilo909:
basically thanks for the plugin.

as SmartHomeNG has been progressed pretty fast and the SmartPlugin is now a central component for years, i would favor a PR with a smartplugin instead of a classic plugin. even now it is hard for us to maintain all plugins and introducing a new "classic" plugin, which will lack of features in upcoming releases, is nearly impossible..

also a plugin specific "plugin.yaml" with metadata is highly recommended. i don't think both will generate a lot of effort, so please consider upgrading

bmxp:
I will try to make this a SmartPlugin after Release 1.6. Until being a SP we will keep the plugin in develop.
Surely we need some tester of the revised plugin. Any volunteers?

PR #210, opened 2019-01-31
Ical plugin: bug fix and updates

onkelandy:
fixed local file handling
code tweaks
plugin.yaml extended


msinn:
- The path to the smarthome directory must not be hardcoded!
- the calender file should not reside directly in the smarthome directory.
   configuration files should be in ../etc and 
   data files should be in ../var/\<specific subdir\>

onkelandy:
I'll have a look and change it.

Something different though. If I want to access a https calendar, sometimes I have problems with SSL verification. Using http URL gives me a moved permanently. My idea was to imlplement a "verify=True/False" to skip verification. That needs to be implemented in the core though. 

Would that be an ok "feature" or is there a better solution?

ohinckel:
> My idea was to imlplement a "verify=True/False" to skip verification.

Maybe you can just use the [network](https://github.com/smarthomeNG/smarthome/blob/develop/lib/network.py#L403) library.

onkelandy:
Oh I'll try that. So the tools.fetch_url is somehow deprecated?

ical/__init__.py
```diff
@@ -43,17 +45,27 @@ def __init__(self, smarthome):
...
             calendars = self.get_parameter_value('calendars')
         except Exception as err:
             self.logger.error(err)
             self._init_complete = False
             return
-
+        try:
+            dir = '{}/var/ical'.format(self.sh.get_basedir())
+            os.makedirs(dir)
+            self.logger.debug('Created ical subfolder in var')
```
ohinckel: We should avoid hard-coded paths. Instead, this should only a default path and the user should be able to configure the cache path in the plugin configuration.

ohinckel: The variable `dir` should be renamed into something other (e.g. `directory` or `dir_path`) to not "clash" with Python's `dir(...)` function.

onkelandy: You mean to set the subfolder name of smarthome/var/ in the plugin.yaml - but not the full path, right?

ohinckel: I would make it possible to configure the *full* path to this directory and only use the directory used above in the code as default.

Generally, I don't like to put (temporary, configuration, ...) files into the application directory. For the default behavior it would be OK to use directories in the application folder. But I would like to have some configuration options to override all these directory defaults.

msinn: @ohinckel what would you think of making the Position of the var Directory configurable in general like the configuration already can be configured to reside outside the application directory? That would include the database-, temp- and further directories residing in the var directory.

onkelandy: I think the default should be defined in the yaml and not in the code..

ohinckel: AFAIK, the `var` directory contains data which should be kept, even if SmartHomeNG will be restarted (like the cache files for items to keep the value when restarting).

So when installing a new version into a new directory I'll loose all this information stored in the `var` directory. The configuration can also be moved to a directory outside `get_basedir()` and I just want to configure the path to an external directory which is not empty when installing a new version.

Another note on `var` directory: we should introduce getter which gets a partial path and will prefix it with the appropriate directory. E.g. `sh.get_vardir('ics')` could return `{base_dir}/var/ics/`. Optionally this method could also take care about creating the directory structure.

The `var` path is hard-coded in the plugin - if we change/rename it we have to adjust the pluging too.

msinn: It is clear to me that plugins ang logging config have to be updated to use something like sh._var_dir (analog to (sh._etc_dir, sh._items_dir, ...) but the number of plugins that use the var directory directly is relatively small. So it could be done.

bmxp: I support the idea of having a getter for the basedir for var. 
For the plugin I think that providing a configurable path is the correct way hwo to do it: 
* a relative path like ``./ics`` that could be appended to the var_dir return by the getter or
* an absolute path like ``/home/smarthome/var/ics`` would be fine, too IMHO

onkelandy: So it would be something like this, depending on the get_vardir function that isn't implemented yet, right?
´´´
config_dir = self.get_parameter_value('directory')
if config_dir.startswith('./'):
  directory = '{}/{}'.format(self.sh.get_vardir(), config_dir.replace('./', ''))
else:
  directory = config_dir
´´´

ohinckel: I was thinking of a function which does both, combines default path or uses given absolute path:
```python
self.sh.get_vardir()                      # returns just '/path/to/sh/var/'
self.sh.get_vardir('ics')                 # returns '/path/to/sh/var/ics/'
self.sh.get_vardir('ics/dir')             # returns '/path/to/sh/var/ics/dir/'
self.sh.get_vardir('/path/to/other/dir')  # returns '/path/to/other/dir/'
```
By using `self.sh.get_vardir(self.get_parameter_value('directory'))` one can
* specify a directory outside the `/var` directory by using a `/` at the beginning
* a directory relativ to the `/var` directory by not starting with a slash

This method can also be used somewhere else when plugins (or the core itself) want to know the `/var` directory or want to work with it. And I think such a method would also be helpful for the `/etc` directory.

onkelandy: Implemented the directory and timeout parameter. The latter needs an update of lib/network.py
The former now works in general - but as soon as the get_vardir is implemented some of the try/except can be removed.

Plugin min version is 1.6 - but then it can't be tested with the develop 1.5

PR #213, opened 2019-03-10
problem running <bound method RTR.stop_controller of <plugins.rtr.RTR object at 0x7fd34ba354e0>>: 'str' object is not callable

ThomasCr:


ThomasCr:
ich erstellte das nochmal neu...

PR #215, opened 2019-03-10
problem running <bound method RTR.stop_controller of <plugins.rtr.RTR object at 0x7fd34ba354e0>>: 'str' object is not callable

ThomasCr:
… object at 0x7fd34ba354e0>>: 'str' object is not callable

### DIFF
--- a/rtr/__init__.py
+++ b/rtr/__init__.py
@@ -328,7 +328,7 @@ class RTR(SmartPlugin):
             if self._controller[c]['stopItems'] is not None and len(self._controller[c]['stopItems']) > 0:
                 for item in self._controller[c]['stopItems']:
                     if item():
-                       if self._controller[c]['actuatorItem']() > 0:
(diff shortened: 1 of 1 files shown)


rtr/__init__.py
```diff
@@ -328,7 +328,7 @@ def stop_Controller(self, item, caller=None, source=None, dest=None):
             if self._controller[c]['stopItems'] is not None and len(self._controller[c]['stopItems']) > 0:
                 for item in self._controller[c]['stopItems']:
                     if item():
-                       if self._controller[c]['actuatorItem']() > 0:
+                       if self._items.return_item(self._controller[c]['actuatorItem'])() > 0:
                            logger.info("rtr: controller {0} stopped, because of item {1}".format(c, item.id()))
                        self._controller[c]['actuatorItem'](0)
```
ohinckel: I think there's one more to fix when setting the value to zero.

ThomasCr:
thanks, fixed it too

PR #216, opened 2019-03-10
Change sh_minversion to 1.5.1

julakali:
As smarthome-ng 1.5b does not exist and appearently, 1.5.1 is not assumed greater than 1.5b, it is currently impossible to use this plugin.

### DIFF
--- a/openweathermap/plugin.yaml
+++ b/openweathermap/plugin.yaml
@@ -13,7 +13,7 @@ plugin:
     support: 'https://knx-user-forum.de/forum/supportforen/smarthome-py/1246998-support-thread-zum-openweathermap-plugin'
     keywords: weather
     version: 1.5.0.3              # Plugin version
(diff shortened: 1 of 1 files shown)


msinn:
The author chose the minimum version by purpose to signal that the plugin is not (or not fully) compatible with v1.5.1 (the current master).

What makes you think otherwise?

The letter signals which level of the develop branch is needed to run the plugin. (By the way: he actual version of the develop branch is 1.5d.)

julakali:
Ok, so i guess i expected a versioning scheme were 1.5.1 > 1.5b (And why wouldn't it?)
Since I'm already bugging you with this, how am I supposed to see the "level of develop branch"? Can't find it in the source.

The given plugin actually seems to work with 1.5.1 btw.
Feel free to close this, sorry for opening this pull request.

msinn:
If you are running the develop version, you can see the version in the log. Otherwise take a look at bin/shversion.py

PR #218, opened 2019-03-12
openweathermap: Add support for forecast data

julakali:
This allows the openweathermap plugin to use the free API for forecasts.
Forecasts are received as 40 distinct 3-hours timeslices.
To provide daily forecasts, the average/min/max of those 3-hours slices is calculated.

The forecast is matched by using "forecast/x/" where x is between 0-39 and indicates the index of the 3-hour prediction from the API.

Daily forecasts are matched by using "forecast/daily/x" where x is "days-in-future - 1", so it ranges from 0-4. (We don't have complete forecast data for the current day)

Up for discussion:
- Use daily index 0 for current day, even if it is possible we have NO prediction (e.g. if forecast was requested at 22.00 the first prediction is already for the next day)?
- Is it okay for you to assume that "the temperature" is the average of all available 3-hour predictions?
- While calculating an average temperature makes (some) sense, average wind direction may not
- The fifths day (daily/4) will not have a complete forecast, as the last timeslice available will be at the current time +4 days

This commit currently includes commented-out debug prints that i will of course remove when we're ready.

### DIFF
--- a/openweathermap/README.md
+++ b/openweathermap/README.md
@@ -138,6 +138,149 @@ owm:
         uvi_date:
             type: num
             owm_matchstring@home: uvi_date
+
+        forecast_3hours: # next 3 hours, use 0-39 for further forecasts
+            time:
+                type: num
+                owm_matchstring@home: forecast/1/dt
+
+            conditions:
+                type: list
+                owm_matchstring@home: weather
+
+            temp:
+                type: num
+                owm_matchstring@home: forecast/1/main/temp
+
+            temp_min:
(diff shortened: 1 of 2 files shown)


julakali:
Is there anything else I can do or does anyone want to discuss this solution?

msinn:
You have to be patient. The maintainer of the plugin is on holyday.

psilo909:
i check it the next days.. i am back for 1 week but i am completely blocked by work and private topics

PR #223, opened 2019-03-21
shjq - a generic json plugin based on 'jq'

t3r:
Hi,

this a simple but efficient plugin to feed data from any json source into smarthome items. Json data gets parsed by the powerful 'jq'.

Branched off of master but should also apply nicely against any other branch as it is completely new.

Hope you like it, I've been using it locally for a couple of weeks now to read data from a SolarWatt EnergyManager, a MyReserve battery and a Fronius inverter.


shjq/plugin.yaml
```diff
@@ -0,0 +1,40 @@
...
+parameters:
+    # Definition of parameters to be configured in etc/plugin.yaml
+    url:
+        type: str
+        mandatory: True
+        description:
+            de: 'URL of the json input (e.g. http://foo.bar/input.json or file:///path/to/data.json)'
+            en: 'URL of the json input (e.g. http://foo.bar/input.json or file:///path/to/data.json)'
+    cycle:
+        type: num
```
ohinckel: The type `int` fits better here. `num` is for floats.

ohinckel:
This seems to be a powerful plugin when dealing with JSON data. To make it more useable for other users it requires some more examples I guess. But for a starting point it's OK.

The only think I don't like is the name `shjq`. It's for sure for SmartHomeNG, so the prefix `sh` is not really required here. But `jq` seems to tell us nothing about the plugin.

My suggestions:
* `json` - just a generic name, plugin is for dealing with JSON data
* `json_jq` - JSON data in conjunction with jq, which is the way how to query the data for items
* `data_json` - a generic JSON data plugin (in case we'll have some others in the future, e.g. `data_xml` which could be a plugin to deal with XML data)

shjq/README.md
```diff
@@ -0,0 +1,114 @@
...
+#### url
+
+The address of the json data. Currently http://, https:// and file:// schemes are supported.
+Note: using absolute file paths for the file:// schema yields in tripple forward slashes. This
+might look funny but it is not an error.
+
+#### cycle
+
+The repetitive polling interval in seconds. Defaults to 30 seconds.
+
```
ohinckel: I think we should put the documentation about the items into the `plugin.yaml` not directly in the `README.md` file. Examples and more details about how to use the plugin should be part of `README.md`.

t3r:
Thanks for the valuable feedback. Except for the plugin name, I have implemented the necessary changes.

For the plugin name, what about "readjson"?
It describes exactly what it does. And should also work for the item-attribute prefix.

ohinckel:
Thanks for updating this so fast! I just suggests some names and I'm not really sure what fits better. Maybe the others here have also ideas/opinions about the name. We also can just leave it as is.

shjq/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def __init__(self, sh, *args, **kwargs):
+        """
+        Initializes the plugin
+        @param url: URL of the json data to fetch
+        @param cycle: the polling interval in seconds
+        """
+        self.logger = logging.getLogger(__name__)
+        self._url = self.get_parameter_value('url')
+        self._cycle = int(self.get_parameter_value('cycle'))
```
msinn: Hint: Since **cycle** is defined as int in plugin.yaml, get_parameter_value returns an int. You don't need to convert it any more.

t3r:
Good catch, that int() slipped through the cracks.
FWIW, I have renamed the plugin to jsonread and the item attribute to jsonread_filter. Both now pretty much describe what they do.

msinn:
I think jsonread is a good name for the plugin.

You should adjust the requirements of the plugin: Remove the requests entries.

Requests is required (in a higher version) by SmartHomeNG by default. The higher version is required because of CVE-2018-18074.

t3r:
Done - requests requirement is gone.

msinn:
I just saw that the PR was against the master branch and reverted it.

@t3r Could you recreate the PR against the develop branch?

PR #224, opened 2019-03-23
DPT 18.001 implemented and tested

henfri:
DPT 18.001 implemented and tested.
Documentation to follow


msinn:
Documentation is missing. The metadata in plugin.yaml has to be updated to reflect the change. The user documentation of plugins is generated from the metadata in the plugin.yaml of the plugin.

henfri:
done, see dedicated PR

PR #227, opened 2019-03-23
Update plugin.yaml

henfri:
dpt 18.001 added


ohinckel:
Btw. you do not need to create a new branch for every change/commit and create a new PR. You can just commit to an existing branch and push the changes again to your remote repository to update the existing PR for this branch. 

This would help us to manage all the PR easily and we do not need to check different PRs for one feature.

msinn:
Ich korrigiere noch die Beschreibung in der plugin.yaml:

Capto_Capture 2019-03-24_01-09-23_PM

Die Beschreibung zum Speichern ist falsch. Speichern geht mit 129-192

henfri:
@msinn: Danke
@ohinckel: Thanks, I will try. The problem was, that I was always 'editing' files from the main (sh.py) repository. Instead, after the first edit, github creates a branch in my repository. For other files, I should then edit the files in *this* branch instead editing them in the main repository (which would create another branch)

PR #230, opened 2019-03-27
Jsonread - read json and filter through JQ

t3r:
Here is a new pull request against develop and after rebasing my commits on develop.


t3r:
Thanks for merging!

PR #238, opened 2019-05-13
Rework of Art-Net Plugin

jentz1986:
- Updated Plugin with Item-capability
- Added feature to update DMX-channels every n seconds
- Added Web-Interface for backend/admin Interface
- Enabled multiple instances
- Updated documentation

Tested with https://www.ulrichradig.de/home/index.php/dmx/art-net-box in a one instance setup with a second instance configured pointing to an IP that did not exist. (DMX Hardware: 1 4channel-dimmer and 1 RGB Spot). Environment: Synology NAS, Docker Container henfri/smarthome-ng:1.60


artnet/__init__.py
```diff
@@ -60,6 +166,17 @@ def __call__(self, var1=None, var2=None):
         if type(var1) == list and type(var2) == type(None):
             self.send_frame(var1)
 
+    def _update_loop(self):
+        if not self.alive:
+            return
+        self.logger.debug("Update Loop")
```
ohinckel: Currently the update method is empty. So in case someone configured a update cycle nothing will happen here.

artnet/plugin.yaml
```diff
@@ -35,36 +60,12 @@ parameters:
...
-
-item_structs: NONE
-  # Definition of item-structure templates for this plugin
-
-logic_parameters: NONE
-    # Definition of logic parameters defined by this plugin
-
-plugin_functions: NONE
-    # Definition of function interface of the plugin
-
```
ohinckel: I think the attributes listed above (`item_structs`, `logic_parameters`, `plugin_functions`) should not be removed.

artnet/plugin.yaml
```diff
@@ -5,21 +5,46 @@ plugin:
...
-    state: qa-passed
+    state: ready
     keywords: dmx
 #    documentation: https://github.com/smarthomeNG/plugins/blob/develop/mqtt/README.md        # url of documentation (wiki) page
 
-    version: 1.3.0                 # Plugin version
-    sh_minversion: 1.3             # minimum shNG version to use this plugin
-#    sh_maxversion:                 # maximum shNG version to use this plugin (leave empty if latest)
-    multi_instance: False
-    restartable: unknown
```
ohinckel: I think the `restartable` setting should be still listed there.

artnet/__init__.py
```diff
@@ -25,32 +26,137 @@
...
-        self.logger.debug("Init ArtNet Plugin done")
+        self.init_webinterface()
+
+        self.logger.debug("Init ArtNet Plugin for %s done" % self._model._instance_name)
 
+    def parse_item(self, item):
+        # items bound to this artnet-subnet
+        if self.has_iattr(item.conf, self.ADDR_ATTR):
+            adr = int(self.get_iattr_value(item.conf, self.ADDR_ATTR))
+            if adr > 0 and adr < 513:
```
ohinckel: I think this condition can be move into the `plugin.yaml` by using `valid_min` & `valid_max` (see [docs](https://www.smarthomeng.de/developer/metadata/item_attributes.html)).

artnet/__init__.py
jentz1986: Ooops, will do that asap.

artnet/plugin.yaml
jentz1986: OK, I based this on the information from the AVM plugin. no issue, this can be fixed.

artnet/plugin.yaml
jentz1986: OK, what are criteria for restartable plugins? Data is gone once plugin is restarted, but rest is stateless...

artnet/plugin.yaml
psilo909: plugin_function is a definition currently used for autocomplete in the logics editor.
item_attributes are for future version of the core.

NONE should be correct here. do i miss sth?

artnet/__init__.py
jentz1986: Is this check based on yaml enforced? If so: fine! If not: it should stay here (in addition to the definition in the yaml).

artnet/plugin.yaml
psilo909: check if your plugin will restart save after stop and then run method is called. then it can be set to True

artnet/__init__.py
jentz1986: Was added to the yaml...

artnet/plugin.yaml
jentz1986: Guess I will implement and test the restarting feature. Was set to true in the YAML.

artnet/__init__.py
psilo909: i think u can then remove the check from your code. it should be directly checked by the yaml information

artnet/plugin.yaml
ohinckel: This flag doesn't tell someone if a plugin restarts in the same state, it just tells you if the plugin can be restarted (some plugins has dependencies or even no code implemented to be restartable, in this case the flag is `False`).

But for this plugin it should be easy implement stop and start methods (just close the connection and reopen it again). If you want to restore the state you could also initialize the value based on the item's current values.

artnet/__init__.py
jentz1986: Is this check as well done once the yaml is edited with a normal text-editor? My code is throwing an error if I try to use address 123456 - not a general validation engine.

jentz1986:
I have updated the code with a read-back routine. Can you tell me why Travis fails here?

artnet/__init__.py
psilo909: i think shng needs to be restarted for that. did that fix it?

psilo909:
> I have updated the code with a read-back routine. Can you tell me why Travis fails here?

i updated the plugins dev branch with a new travis.yaml, which worked for the core to fix it.. hope it will do so here also

jentz1986:
I found some issues in my code. Let me fix them first before you integrate this...

jentz1986:
OK from my end now...

artnet/__init__.py
```diff
@@ -21,36 +22,170 @@
...
+                                   net=int(self.get_parameter_value(
+                                       'artnet_net') or 0),
+                                   subnet=int(self.get_parameter_value(
+                                       'artnet_subnet') or 0),
+                                   universe=int(self.get_parameter_value(
+                                       'artnet_universe') or 0),
+                                   instance_name=self.get_instance_name(),
+                                   update_cycle=int(
+                                       self.get_parameter_value('update_cycle') or 0),
+                                   )
```
ohinckel: All these `get_parameter_value(...) or XY` are use to get the parameter value or a default value? The default value for a parameter can also be specified in the `plugin.yml`.

jentz1986: Yes but this is neither in effect if you change the installations plugin.yaml manually nor if you have shng151, right? I added the defaults to the plugin.yaml (of this plugin) now.

jentz1986:
Is there any checklist for "what to check yourself before raising a pull-request"?

artnet/__init__.py
msinn: You are wrong. Since shng 1.4 the get_parameter_value(...) returns the default value, if no value is configured for the parameter in /etc/plugin.yaml. You can even simplify it further. You don't need to convert the values to int. get_parameter_value(...) returns the value in the datatype that corresponds to the type defined in plugin.yaml (as described here: https://www.smarthomeng.de/developer/development_plugin/smartplugin.html?highlight=get_parameter_value#lib.model.smartplugin.SmartPlugin.get_parameter_value)

artnet/__init__.py
```diff
@@ -89,9 +89,21 @@ def get_items(self):
...
         Returns added items
 
-        :return: array of items held by the device
+        :return: array of items held by the device, sorted by their DMX-address
+        """
+        instance_key = "artnet_address"
+        if self._instance_name:
+            instance_key = "artnet_address@"+self._instance_name
+
+        return sorted(self._items, key=lambda i: i.conf[instance_key])
```
ohinckel: To access the item's configuration you should use the `plugin.get_iattr_value(i.conf, 'artnet_address')`. This will take care about the instance name. When changing the logic how we handle instance names in configuration your solution will break.

artnet/plugin.yaml
```diff
@@ -61,14 +63,22 @@ parameters:
...
         valid_min: 0
         valid_max: 15
+    min_channels:
+        type: int
+        default: 1
+        description:
+            de: 'Gibt an, wie viele Kanäle mindestens gesendet werden müssen, hilfreich für Geräte, die einen vollen Datensatz über alle ihre Kanäle erhalten müssen'
+            en: 'Specifies how many DMX channels are to be sent at minimum'
+        valid_min: 1
+        valid_max: 512
```
ohinckel: What the reason for a min-channel settings. As the code tells me, it's only used to fillup the data buffer which gets send to the DMX network. Before it was dynamically increased depending on value changes to the addresses: if the data buffer does not contain a value for the given address, the buffer will be increased up to the given address and the new value will be put into the data buffer.

We could also initialize the complete data buffer for all addresses when parsing items finished. So when the plugin identified 64 as the maximum address the plugin could initialize the data buffer with 64 values (initial value set to zero). When now sending the data to the DMX network it will always send the 64 values to it. No need to specify an explicit value for minimal channels to send.

Maybe I oversee something, but I can't imagine a benefit on having such an attribute when we also could have a default initialization of the data buffer with the maximum address used in configuration.

artnet/__init__.py
jentz1986: Yes, after now having gained more experience with writing PlugIns I will do so...

artnet/plugin.yaml
jentz1986: This is to avoid issues as described here:
https://knx-user-forum.de/forum/supportforen/smarthome-py/996278-problem-mit-artnet-plugin-bzw-der-dmx-hardware

Imagine a setup as follows: 
- One 36-channel dimmer (to be obtained easily and quite cheap). 
- Only e.g. 32 channels are used and the 33rd till 36th channel are left unconnected. Economic solution as there is no cheaper way to control 8 rgbw lights, e.g. in a garden-setup.
- Only items are specified, no logic is run.

So if you send "all LEDs white" no light will light up as the dimmer requires all 36 channels to be defined. To avoid this you can use this config to ensure the transmission of at least n channels. Otherwise ignore it.

artnet/__init__.py
jentz1986: solved it

artnet/__init__.py
jentz1986: fixed it accordingly

onkelandy:
@jentz1986 can you please try to resolve the conflicts?

jentz1986:
Unfortunately I have no write access to this repository. 

So what is the issue here:
When I started unifi development I forked from development (which already contained the artnet-change). I the reverted the artnet-change within the unifi-fork. The unifi fork has been integrated to development (so my artnet-change was applied and rolled-back). Now there are three versions: The original version, the rolled-back original version and the new development. The new development is now in my repository, but is older than the revert from above.

How can I contribute best now? Shall I fake change things (e.g. add empty lines) in my repo? Or is it possible to just say "resolve by using head of develop branch on jentz1986/smarthomeNG-plugins."? The latter operation is not available to me...

ohinckel:
I would suggest to do a new checkout of your [develop branch](https://github.com/jentz1986/smarthomeNG-plugins/tree/develop) from GitHub and create a new branch `artnet-rework`. Then add a new remote to the [forked repository](https://github.com/smarthomeNG/plugins) and fetch the `develop` branch. After this you should be able to do a rebase in your `artnet-rework` branch with the fetched develop which will end up in the conflict state. After fixing the conflict you can continue your rebase (and potentially end up in the next conflict). When all conflicts are resolved you have a version which is now based on the current develop branch.

```
git clone https://github.com/jentz1986/smarthomeNG-plugins.git && cd smarthomeNG-plugins
git checkout develop
git checkout -b artnet-rework
git remote add vendor https://github.com/smarthomeNG/plugins.git
git fetch vendor develop
git rebase vendor/develop
[... solve conflicts, after this continue with: git rebase --continue ...]
```
After this check if your changes are in the current develop branch you're working in. If everything is OK you can push your changes to your remote repository on GitHub.

```
git push origin artnet-rework
```
This will create the new branch on GitHub too. This can be used to create a new PR.

If you're sure you're doing everything right, you can also work directly in your develop branch instead of creating a new branch `artnet-rework`. You only need to use `git push --force` to push your changes to GitHub because your history changed and is based on different commit (hopefully you don't have any other branches which are based on old versions).

jentz1986:
This is a duplicate of #251 now. Whereas #251 contains the correct non-conflicting code.

PR #245, opened 2019-05-29
Develop unifi plugin

jentz1986:


jentz1986:
> Will you really need a .gitignore with so many entries at the plugin level?
Hmm, just copied the template and didn't think about it anymore...

onkelandy:
I got this error at startup.. maybe my item config is not the way it was intended to be ;) Could you please provide an example item tree. Aa a struct template in the plugin.yaml if possible.

```
2019-06-02 23:37:45 CEST ERROR    lib.item            item                __init__     ln:685 Item unifi.switch_port_enabled: problem creating: 'Items' object has no attribute 'path'  ----  Main         
Traceback (most recent call last):
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 357, in _get_attribute_recursive
    if item_type in item.conf:
AttributeError: 'Items' object has no attribute 'conf'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/ank/smarthome/lib/item.py", line 683, in __init__
    child = Item(smarthome, self, child_path, value)
  File "/Users/ank/smarthome/lib/item.py", line 748, in __init__
    update = plugin.parse_item(self)
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 445, in parse_item
    prof_on = self._get_attribute_recursive(item, UniFiConst.ATTR_SW_PORT_PROF_ON)
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 369, in _get_attribute_recursive
    return self._get_attribute_recursive(item.return_parent(), item_type, check, leaf_item)
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 369, in _get_attribute_recursive
    return self._get_attribute_recursive(item.return_parent(), item_type, check, leaf_item)
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 367, in _get_attribute_recursive
    item_type, leaf_item.path()))
AttributeError: 'Items' object has no attribute 'path'
```

jentz1986:
Whoa, how did you manage to get that? Can I see your items.yaml? I thought I had covered all cases like
item not there, item has no settings, item is somehow set wrong (mac). I would like to cover your case as well...

However: There is an example items.yaml in the readme.md

onkelandy:
Ooops, didn't see the README ;) Makes things easier.
Nevertheless the error handling could be improved for sure. If the hierarchy isn't correct the above error will be there. So if you have a client_present_at_ap but there is no ap above it. Makes sense to throw an error there for sure so it was mostly my mistake.

However, this results in the same error:
```
  switch:
    unifi_switch_mac: xx:xx:xx

    ap_eg:
      unifi_ap_mac: xx:xx:xx
      unifi_switch_port_no: 5

      port_enabled:
        type: bool
        unifi_type: switch_port_enabled
```

The rest seems to work pretty well. Thanks for your work! Cool stuff!

P.S.: Personally I prefer a nice and clean error message if something is wrong instead of the whole raiseError thingy. Keeps the log cleaner and makes parsing/automatic evaluation easier.

onkelandy:
What's somehow not clear to me.. what are devices? In the Unifi Dashbaord they distinguish between Switch, AP, Clients and Guests. Would be cool to have the same in the plugin. They do indeed have devices, but there are AP and switches "combined" so it's not really useful imho. 

In general I'm wondering if it wouldn't be easier to keep the attribute name variety to a minimum. Instead of all the different MAC and client/device attributes maybe an additional device_type attribute?

Something like:
```
switch:
  unifi_mac: af:fe:af:fe:00:00
  unifi_device_type: switch
  
  ip:
    type: str
    unifi_function: ip
    unifi_device_type: switch #might not be necessary as the type is already defined in the parent item

  sw_name:
    type: str
    unifi_function: name
  
  access_points:

    ap_ac_lite1:
      unifi_device_type: ap
      unifi_mac: af:fe:af:fe:00:01
      unifi_switch_port_no: 5

      ap_ip:
        type: str
        unifi_function: ip

      ap_name:
        type: str
        unifi_function: name
      
      ap_enabled:
        type: bool
        unifi_function: enabled

```

jentz1986:
> Thanks for your work! Cool stuff!
Thanks for the flowers! Feels good to not be confronted with why is the file like this or that (things that could easily be avoided if one would know about the need to check)

> P.S.: Personally I prefer a nice and clean error message if something is wrong instead of the whole raiseError thingy. Keeps the log cleaner and makes parsing/automatic evaluation easier.
Absolutely, but unfortunately if one is inside the things one is tending to overlook issues or underestimate the users creativity... I want to improve the error-handling, therefor I had attached the "item-issues" in the web-page. I will take your example to improve on that topic.

Besides the item-generator on the web-interface should create a valid configuration based on the actual available controller data. Unfortunately I have not managed to figure out a way to sort the output properly...

Regarding your item-type suggestion:
I was thinking the same way. But then I figured out to be more efficient with the separate types. However your other idea would make sense to distinct ap_ip and switch_ip instead of device_ip.

Foxi352:
I also tested it this evening. No Unifi switch. Only 3 UniFi AP-AC-Pro.
Controller version 5.6.26 running on Debian Jessie.

Works great, thank you :-)

jentz1986:
Thank you Serge / Foxi352!

@onkelandy : Found the issue. The Error-handling had a bug (wrong number of parameters in recursion). I want to improve the topic to include the possibility to have a default setting in place if the port-profile cannot be found. But my course is now starting so I have to shift it to the evening...

onkelandy:
Error handling looks good now, though the logger doesn't give any information about wrong configs. When you have a client_present_at_ap but the item is not parented to an ap I don't get any info on that.

WebIF only delivers Error 500 - even with no unifi items or just one switch. Don't know how to debug the webif sorry :(

Foxi352:
> Error handling looks good now, though the logger doesn't give any information about wrong configs. When you have a client_present_at_ap but the item is not parented to an ap I don't get any info on that.
> 
> WebIF only delivers Error 500 - even with no unifi items or just one switch. Don't know how to debug the webif sorry :(

I confirm error 500 on webif when no unifi items at all.

jentz1986:
With the two latest versions I could not reproduce the 500 with no items. I have added the exception output now to the webinterface area for the item-generator as I suspect the problem there.

onkelandy:
Looking pretty good now! Thanks!
Nevertheless what is misleading are the "problem" entries in the webIF. Even if they are white and "ok".. they shouldn't be called problems/issues.

Whatever the item generator should do - I get a list index out of range

jentz1986:
The issue with the Item-generator is that it connects to your controller, reads your infrastructure and populates a full-blown items.yaml with your infrastructure. In my case (four mobiles, six PCs, three smart devices, two UAP and one USW) this generates more than 400 lines of yaml. That would be needed if one would like to reflect the UAP, USW and the clients within shng. Idea is: Take it, copy it to the editor of your choice, remove what you don't like, save it to your structure or as extra-file, done.

Can you spot the point (line of code) where the out of range occurs? Issue is that I only have my setup. I can try to fake data output from the controller, but this could as invalid as my code thinks your output is...

onkelandy:
Cool idea!!! Had to dig deep but found the problem. After changing line 180 in the code to
```
node_body[single_dev_key]['port_enabled'][UniFiConst.ATTR_SW_PORT_PROF_ON] = self._api.get_port_profile_for(
                        n['mac'], port_no-1)
```
it worked for me. The port_profile was indeed offset one port every time before. I have an 8 switch port and the code "crashed" at port 8 all the time.

onkelandy:
I've now copied the output of the item generator and have some other issues now.

The first one is pretty obvious in line 613 ;)
```
self._log_item_warning(item, "Current port-profile \{}}\"
```

Line 591 also has to be changed:
```
return self._model._api.get_port_profile_for(sw_mac, sw_port - 1)
```

Now everything seems to run. Nice!!!!! Some further suggestions though.
In the webIF it might be useful to see the unifi_type because it's not clear what bool True means. Could be anything (right now it's present. But also could be present_at_ap).

What do you think of two buttons in the "Item Generator" page:
- generate standard items
- generate extended items
With a short description that the latter also creates "present_at_ap" entries for all wireless devices in combination with all ap's?

jentz1986:
I would like to check whether the wrong port-numbering can be fixed in the api instead, so the plugin stays cleaner. My time this weekend is a bit limited but I will try.
I have started with a shift of the problems tab to the items itself. Made good progress, but need some time.
Instead of the two buttons my idea went to the direction to have a kind of assistant: The following unifi devices are found, include them (checkboxes)? These wired clients exist, check those, that you would like to check for presence. Regarding the wireless clients: a list of all clients, one column with a checkbox for each access-point and one for “any” ap. 
In this case YAML would be JavaScript generated - solving the sorting issue as well...

Focus: bugfix for port-numbering, then killing the problems tab.

jentz1986:
@onkelandy 
Sorry to say so, but I could not find any issue with the code of the port-profile +1/-1 I have assigned specific port-profiles to each of my ports and the get assigned well for all eight ports of my switch. So there has to be an issue at some other place. Definitely as it is working with -1 on your end there _has_ to be some difference in our environments.
Can you save your fixed version and check my new one as well? I have improved error-handling even more and have tweaked the UI a bit. Please do not skip the ubiquiti/unifi.py update. The code is from me as well. (in this update I only removed some commented line, i know).
In my environment the USG is now connected to port 8 so no AP there...

onkelandy:
With your new code (exchanged the whole unifi folder) the item generator runs without problems. But still the profiles are assigned wrongly. 

I found out what the problem is, though!!! My ports 3 and 4 are aggregated! Sorry I didn't notice before - that's for sure the root of all evil.

I now also get this error by the way:
```
2019-06-16  13:24:18 ERROR    plugins.unifi.poll_unifi Method plugins.unifi.poll_unifi exception: list index out of range
Traceback (most recent call last):
  File "/Users/ank/smarthome/lib/scheduler.py", line 522, in _task
    obj()
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 658, in poll_device
    self._poll_with_unifi_type(UniFiConst.TYPE_SW_PORT_ENABLED, lambda i: self._check_sw_port_enabled(i))
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 646, in _poll_with_unifi_type
    item(func(item), self.get_shortname())
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 658, in <lambda>
    self._poll_with_unifi_type(UniFiConst.TYPE_SW_PORT_ENABLED, lambda i: self._check_sw_port_enabled(i))
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 619, in _check_sw_port_enabled
    rslt = self._get_sw_port_profile(item)
  File "/Users/ank/smarthome/plugins/unifi/__init__.py", line 606, in _get_sw_port_profile
    return self._model._api.get_port_profile_for(sw_mac, sw_port)
  File "/Users/ank/smarthome/plugins/unifi/ubiquiti/unifi.py", line 305, in get_port_profile_for
    port_prof = poData[port_number - 1]['portconf_id']
IndexError: list index out of range
```

onkelandy:
Looking pretty cool now. I guess the warning here is a result of the aggregated port?
```
Unable to determine current switch-profile for switch xxx in item unifi_network.devices.usw_unifi_switch.wired_client_synology.port_enabled


```

Small issue for the webif: translate: Text 'Uni-Typ' to language 'de' -> no translation ('de' or 'en') found

jentz1986:
I have EXACTLY the same ports aggregated. 😄 But at the time of first coding I hadn't. It worked in both cases, before and after. port 4 is still in the list of data that is returned by the controller. I have now changed the algorithm to find the port-profile to not check the index, but the port-idx given by the controller, both when reading and writing. Guess your config is clean.

Fixed:  
- Text 'Uni-Typ' to language 'de' -> no translation ('de' or 'en') found
- Exception with no network connectivity during startup (I hope - can't test that)
- Error-handling improved during polling and item-update

onkelandy:
Still got the errors when polling. It can be fixed by putting the scheduler line here:
```
try:
            self.poll_device()
            self.scheduler_add('poll_unifi', self.poll_device, cycle=self._cycle)
```

Nevertheless - the plugin now leads to some freeze on stopping smarthomeNG (and having no connection). The process has to be killed and stopping takes much longer. I couldn't figure how to fix the problem. Maybe you could run shng locally on a computer with screen/keyboard attached and network disabled?

onkelandy:
Can you please have a look at the item generator problem when devices have a " " in their name? I guess we are very close to a merge then ;)

psilo909:
from my side all ok so far..

onkelandy:
@psilo909 Do you have Devices with blanks in the name and did you try the item generator?

psilo909:
@onkelandy no, thats why i did not say "this issue is working here" ;)

onkelandy:
So I guess it's the last issue before merging.. ;)?

jentz1986:
If I wouldn't forget to push it would have been available earlier... Damn.

jentz1986:
> just VERY VERY minor: you can add me as tester in plugin.yaml

thank you! Done that. Question: When to set the state to qa-passed? Or is this done after merging? And: Currently no logic functions are exposed. I guess it is OK, or do I miss something?

onkelandy:
The blank replacement also needs to happen for wlc_key = please (line 245)

onkelandy:
Dann würde ich sagen, mergen wir mal. Echt super geworden, das Plugin. Vielen Dank @jentz1986 !!

PR #259, opened 2019-07-18
sonos Plugin: Adaptation to the version SoCo 0.17 (including changes from pfitschi)

ivan73:


aschwith:
I tested this pull request with the latest Sonos revision 10.3 and can confirm that it fixes the problems with transport actions that occurred recently. Also, I positively tested the TTS feature. I have not checked whether the SoCo memory leak still exists and will refer back to that after a while of monitoring.

PR #269, opened 2019-08-07
snap7_logo plugin: new plugin

ivan73:
new plugin to communicate with the Siemens logo using the snap-7 library.


msinn:
I haven't looked at the complete plugin yet, but here is a first finding:

The plugin parameters (and item attributes) should be defined with their default values in plugin.yaml instead defining them in the source code. The way it is done now, it is not possible to configure the plugin through the admin interface and SmartHomeNG can't do any type checking while loading the parameters.

ivan73:
I have added the parameters in the plugin.yaml. Do I have to make a new pull request now?

msinn:
As you can see in this conversation, the two commits you mad to your repo have automatically been added to this PR. You don't need to make a new pull request.

msinn:
I found the following:

- ```self.logger = logging.getLogger(__name__)``` should be removed. It is handled in the SmartPlugin Class with the extension that the instance name is logged when configuring multiple instances on MI plugins. If you define the logger yourself, the instance name isn't logged automatically.

- In the line following the logging.getlogger() the indentation is wrong.

PR #273, opened 2019-08-11
plugin telegram: added new attribute 'control' to change (toggle) sh-item-value (bool)

ivan73:


telegram/__init__.py
```diff
@@ -215,6 +219,28 @@ def parse_item(self, item):
             if value in ['true', 'True', '1']:
                 self._items_text_message.append(item)
             return self.update_item
+        
+        if self.has_iattr(item.conf, ITEM_ATTR_CONTROL):
+            attr = self.get_iattr_value(item.conf, ITEM_ATTR_CONTROL)
+            #self.logger.debug("parse item: {0} {1}".format(item, key))
+            name = item.id()    # default
+            type = 'toggle'     # default
```
bmxp: `type` can be used to check the type of a python object. Although there is another function isinstance() you should not use type here.

telegram/__init__.py
```diff
@@ -215,6 +219,28 @@ def parse_item(self, item):
...
+            par_list = attr.split(',')  # Parameter from attr example: 'name:test, type:toggle'
```
bmxp: Shouldn't attr now be a list of strings here?

telegram/plugin.yaml
```diff
@@ -109,6 +109,12 @@ item_attributes:
             de: 'Ein Item dessen Datentyp dict ist und Paare von ``Chat_id`` der vertraut wird und Schreibzugriff ``True`` oder ``1``'
             en: 'An item of type dict with trusted ``chat_id`` and write access ``True`` or ``1``'
 
+    telegram_control:
+        type: str
```
bmxp: Type string does not seem correct here. It looks as if it was a list of strings from the sample above.

PR #284, opened 2019-10-06
New Plugin: helios_tcp

freget:
The plugin allows the communication with the newer generation Helios devices over Modbus TCP.


freget:
Sorry for the confusing commit history. I did not keep the two PRs I have open right now cleanly separated from each other and thus had to revert a bit for this branch.

PR #285, opened 2019-10-06
Mailrcv plugin: Added option to change name of trash and fixed a problem occuring when decoding emails

freget:
This PR contains of two small bugfixes:

**Trashfoder option:** 
Depending on the configuration of the IMAP server, the Trash folder can have different names.
To move a read mail to trash, the mailplugin has to know that name. Added the possibility to change
the name of the trashfolder.

If deleting the mail fails a list of abailable folders is written to the log.

**Emails without content-transfer-encoding header**
According to RF1341, emails can but must not contain the content-transfer-encoding header. (https://www.w3.org/Protocols/rfc1341/5_Content-Transfer-Encoding.html).
If this header does not exist, the python email package does not allow decoding of the message (at least in some versions).

To resolve this we simply set the content-transfer-encoding header to 7BIT, which is the default, if it does not exist.

### DIFF
--- a/mailrcv/__init__.py
+++ b/mailrcv/__init__.py
@@ -31,7 +31,7 @@ class IMAP(SmartPlugin):
     ALLOW_MULTIINSTANCE = True
     PLUGIN_VERSION = "1.4.0"
 
-    def __init__(self, smarthome, host, username, password, cycle=300, port=993, tls=True):
+    def __init__(self, smarthome, host, username, password, cycle=300, port=993, tls=True, trashfolder="Trash"):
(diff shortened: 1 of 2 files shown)


bmxp:
Thank you for your PR. I just have two remarks:
1. you should probably increase the version number minor
2. the __init__ parameters should be changed to the new format like:

```python
    def __init__(self, sh, *args, **kwargs):
        self._host= self.get_parameter_value('host')
        self._port= self.get_parameter_value('port')
```
etc.

freget:
I fixed those two points. Please review again.

PR #293, opened 2019-11-01
Fix class name

robert-budde:
class_name should be written always the same - leads to confusion if it is sometimes spelled classname while meaning the same thing. Best example: https://knx-user-forum.de/forum/supportforen/smarthome-py/1422178-class_name-und-classname-konsolidierung-der-schreibweisen - this has to be merged together with the changes in the smarthome repo!

### DIFF
--- a/alexa/plugin.yaml
+++ b/alexa/plugin.yaml
@@ -16,7 +16,7 @@ plugin:
     sh_minversion: 1.3             # minimum shNG version to use this plugin
 #    sh_maxversion:                 # maximum shNG version to use this plugin (leave empty if latest)
     multi_instance: False          # plugin supports multi instance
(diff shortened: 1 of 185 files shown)


robert-budde:
see https://github.com/smarthomeNG/smarthome/pull/334

robert-budde:
instance_name etc is also spelled like class_name - another one would be "classpath" -> "class_path".

msinn:
I can't review a PR that changed 185 files and be sure it would not break any functionality. Especially since it is not bringing or fixing any functionality. It is fixing cosmetic things only.

Don't get me wrong, i don't propose ignoring programming guidelines, but a find the risk of this PR is too high only to enforce programming guidelines.

PR #320, opened 2020-01-15
add Xiaomi Robvac Plugin

Bonze255:


bmxp:
@Bonze255 Would you please set up a PR against develop? We won't touch the master branch unless a new release comes out or there are urgent bugs to fix.

PR #341, opened 2020-04-07
Plugin: Yamahayxc - update to 1.0.6

Morg42:
added functionality for alarm clock settings (on supported devices)
minor fixes
included item struct functions require sh.py 1.7 (nested structs)

### DIFF
--- a/yamahayxc/plugin.yaml
+++ b/yamahayxc/plugin.yaml
@@ -18,6 +18,7 @@ plugin:
                 - Wiedergabesteuerung (Play/Pause/Stop/nächster,vorheriger Titel/Vor-,Zurückspulen),\n
                 - Lautstärkesteuerung\n,
                 - Anwahl von Favoriten und Presets (geräteabhängig)\n
+                - Stellen und De-/Aktivieren des Weckers (geräteabhängig)\n
             \n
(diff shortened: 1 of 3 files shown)


yamahayxc/plugin.yaml
```diff
@@ -61,16 +63,116 @@ plugin:
 #    documentation: https://github.com/smarthomeNG/smarthome/wiki/CLI-Plugin        # url of documentation (wiki) page
     support: https://knx-user-forum.de/forum/supportforen/smarthome-py/1174064-plugin-yamaha-musiccast-geräte-neuere-generation
 
-    version: 1.0.5                 # Plugin version
-    sh_minversion: 1.3             # minimum shNG version to use this plugin
+    version: 1.0.6                 # Plugin version
+    sh_minversion: 1.7             # minimum shNG version to use this plugin
```
bmxp, line 67: Which breaking change leads to incompatibility to older SHNG versions?

Morg42: No breaking change as far as I know. Nested structs only work with 1.7, structs earlier. Base code is not changed in respect to core libraries. Reverted minversion to 1.3

PR #353, opened 2020-04-13
Update plugin.yaml

henfri:
Segment status must be a list.
Currently, the error is: 
 Item eg.Wohnzimmer.robo.live.segment_status: value "[1]"


onkelandy:
Thanks. Gonna implement that in the next PR.

PR #354, opened 2020-04-13
Update requirements for AlexaRc4shNG

Andrek01:
Hello there was a bug in uploading the last version, the requirements.txt was still wrong.
i changed the path to the documentation in plugin.yaml, it was also wrong
Sorry


msinn:
Why did you remove the requirements, Don't you need pycurl any more? You added a comment "requests moved to core" but pycurl isn't used in the core.

Andrek01:
changed some typos

PR #358, opened 2020-04-14
Onewire Plugin fix format string in debug output

henfri:
Fixes ValueError: cannot switch from automatic field numbering to manual field specification


onkelandy:
Please rename the PRs so that it's clear what it is about.
Prefix it with "Plugin XYZ:"

PR #359, opened 2020-04-15
xiaomi_vac - enable segment clean

henfri:


onkelandy:
clean_zone and segment_clean in struct

PR #363, opened 2020-04-18
Plugin yamahayxc: updated struct definition, changed state to ready

Morg42:
- updated struct definition with relativ item references, so no further item config is necessary
- changed state to ready


Morg42:
Fixed plugin.yaml.

PR #364, opened 2020-04-21
InfluxDB: HTTP write added

chrisr79:
- added HTTP write 
- workaround for "invalid tag format" error at ":ga"
[influxdb.zip](https://github.com/smarthomeNG/plugins/files/4510330/influxdb.zip)


msinn:
Your PR tries to merge 47 files (containing code from 20 plugins) from the develop branch of SmartHomeNG into the master branch of SmartHomeNG. That makes no sense. Only members of the core team are allowed to make releases of SmartHomeNG.

Please create a new PR for your change to influxdb (which is not contained in the actual PR) the

PR #382, opened 2020-06-11
Ical Plugin: scheduler use smartplugin method

onkelandy:


onkelandy:
> Shouldnt there be a scheduler_remove in the stop method?

fixed here: https://github.com/smarthomeNG/plugins/pull/387

PR #391, opened 2020-06-26
Plugin rpi1wire: update v1.7.0 with webinterface

soft-rudi:
Hallo,
dies ist die neue Version des rpi1wire - Plugin.
Die ToDo - Bemerkungen wurden abgearbeitet, danke dafür, und das Webinterface ist implementiert.


bmxp:
We can not merge this PR because there are way too many commits that have nothing to do with the change for the rpi1wire plugin. Also why are the previous versions deleted, did they have errors that make them unsafe to use?

A suggestion would be to locally checkout a whole new plugin directory and use a tool to copy just the changes from rpi1wire to the new plugin directory and then use this to either create a patch file or make a new PR.

If all fails you can post a zip of the plugin in a gist and we can manually merge it.

soft-rudi:
Ok, I'll try a new fork and close this PR.

PR #402, opened 2020-08-31
squeezebox plugin: switch from connection lib to network lib

onkelandy:


bmxp (changes requested):
Well done. Would it be appropriate to bump up the version to 1.4 since it will be a major change and it will be earliest run with SHNG 1.7 develop so this should be changed, too.
And maybe you can make a subfolder ``_pv_1_3_2`` for the old plugin and put the old version inside? Just to ensure the plugin is stil available for people using dead horses to ride aka old SHNG versions

onkelandy:
> Well done. Would it be appropriate to bump up the version to 1.4 since it will be a major change and it will be earliest run with SHNG 1.7 develop so this should be changed, too.
> And maybe you can make a subfolder `_pv_1_3_2` for the old plugin and put the old version inside? Just to ensure the plugin is stil available for people using dead horses to ride aka old SHNG versions

done

PR #406, opened 2020-09-20
XMPP: Replace sleekxmpp with slixmpp

ohinckel:
The XMPP Python library [sleekxmpp](https://github.com/fritzy/SleekXMPP) is no longer maintained the the repository is archived. There's a replacement library called [slixmpp](https://github.com/poezio/slixmpp).

Changing the XMPP plugin to use the new library (with some adjustments to the methods used). But mostly it seems to be compatible.

### DIFF
--- a/xmpp/README.md
+++ b/xmpp/README.md
@@ -68,6 +68,95 @@ Description of the attributes:
 * use_ipv6: enable IPv6 support, which is the default
 * plugins: list of plugins (XEP to support)
 
+#### XEP supported
+
+- XEP-0004: Data Forms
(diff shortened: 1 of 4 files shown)


bmxp (commented):
I would think it might be a good idea to add your name to the copyright in header and increase the version number by 0.0.1 in ``plugin.yaml`` and ``__init__.py``

bmxp:
According to the KNXUF it seems as if there is no one using this Plugin except @ohinckel. Maybe you want to set yourself als maintainer, create a support thread and fill in the keywords and documentation in plugin.yaml?

PR #407, opened 2020-10-19
bugfix for Web-IF, some improvements

Andrek01:
see headline


bmxp:
If you prepare a PR it would be nice to start with "Plugin alexa4p3  ..."

psilo909 (commented):
> If you prepare a PR it would be nice to start with "Plugin alexa4p3 ..."

the PR covers 2 plugins. it would be better to have separate ones.

Reason for the comment above: we assble the release notes out of the PR titles and Commit titles.

PR #411, opened 2020-11-08
Avm plugin fixes

mptei:
AVM-Plugin: If there is no 'call_duration_incoming' item, then a KeyError happens when trying to access self._duration_item['call_duration_incoming']. This exception is caught, but then when trying to log this error an exception occurs. This second exception is reported to console only. That means nothing is written to the log.

This pull request fixes the logging error and also creates the dictionary entries 'call_duration_outgoing' and 'call_duration_incoming'.

### DIFF
--- a/avm/__init__.py
+++ b/avm/__init__.py
@@ -60,6 +60,8 @@ class MonitoringService:
         self._items_incoming = []  # items for incoming calls
         self._items_outgoing = []  # items for outgoing calls
         self._duration_item = dict()  # 2 items, on for counting the incoming, one for counting the outgoing call duration
+        self._duration_item['call_duration_outgoing'] = None
+        self._duration_item['call_duration_incoming'] = None
         self._call_active = dict()
         self._listen_active = False
         self._call_active['incoming'] = False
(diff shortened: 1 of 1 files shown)


avm/__init__.py
```diff
@@ -298,7 +300,8 @@ def _trigger(self, call_from, call_to, time, callid, event, branch):
                 self._call_incoming_cid = callid
 
                 # reset duration for incoming calls
-                self._duration_item['call_duration_incoming'](0, self._plugin_instance.get_shortname())
+                if not self._duration_item['call_duration_incoming'] is None:
```
psilo909, line 303: Is this really working (did u check it?)? 
Normally I check for the existence of keys in dicts with "if 'call_duration_incoming' in self._duration_item: ..."

So I am not sure if the None comparison works, because I think its trying to access the non existing key before comparing?

psilo909: ok no i see, you are initializing it with None above.. then ok..

psilo909: Hmm I think I would still prefer to fix it with the
if 'call_duration_incoming' in self._duration_item
check...
In my view a non existing key signals better the item does not exist than one with "None" as value.

mptei: I've added the checking for None because this is done often in the code.

psilo909: just checked.. you may be right.. accepting the PR

PR #416, opened 2020-11-10
Plugin bsblan

thengsty:
- added support for writing values to boiler system
- bugfixes
- Rework readme file


bsblan/__init__.py
```diff
@@ -24,8 +24,11 @@
...
 from urllib.request import urlopen
 import json
+import requests
+
 from lib.model.smartplugin import *
 
+
 class Bsblan(SmartPlugin):
     ALLOW_MULTIINSTANCE = False
     PLUGIN_VERSION = '1.0.0'
```
bmxp, line 34: I think the plugin version at this point here should also updated as well as the version information in plugin.yaml

bsblan/__init__.py
```diff
@@ -50,13 +53,59 @@ def parse_item(self, item):
         if 'bsb_lan' in item.conf:
             self._items.append(item)
 
+        if self.has_iattr(item.conf, 'bsb_lan'):
+            return self.update_item
+
     def poll_device(self):
         for item in self._items:
-            self.url = 'http://'+str(self.get_parameter_value('bsblan_ip'))+'/JQ='+str(item.conf['bsb_lan'])
+            self.url = 'http://' + str(self.get_parameter_value('bsblan_ip')) + '/JQ=' + str(item.conf['bsb_lan'])
```
bmxp, line 61: it might be more readable to use something like ``self.url = 'http://{0}/JQ={1}'.format(self.get_parameter_value('bsblan_ip'), item.conf['bsb_lan'])``
(Probably even prettier in the next versions using f-strings)

PR #420, opened 2020-12-02
CLI Plugin: Use items API for all related commands

onkelandy:


bmxp (commented):
You should probably also bump up the version number in ``__init__.py`` and ``plugin.yaml`` and have a look if it is a breaking change for an earlier version of SHNG caused by the items API

onkelandy:
bumped to 1.7.2
sh min version is already 1.7, pretty sure that should work, right?

PR #433, opened 2020-12-28
added support for solarlog firmware >=3.x

cipcm:


solarlog/plugin.yaml
```diff
@@ -21,6 +21,14 @@ plugin:
 
 parameters:
     # Definition of parameters to be configured in etc/plugin.yaml (enter 'parameters: NONE', if section should be empty)
+    fw:
```
bmxp, line 24: Did you check the plugins metadata with ``python3 tools/plugin_metadata_checker.py -c solarlog``
Probably an attribute fw (firmware) could be a list of existing firmware flavours?

solarlog/__init__.py
```diff
@@ -72,7 +77,12 @@ def run(self):
         """
         self.logger.debug("Run method called")
         # setup scheduler for device poll loop
-        self.scheduler_add(self.get_shortname(), self.poll_device, cycle=self.cycle, next=shtime.now())
+        if self.fw:
+            # use 'old' code for firmware <= 2.x
+            self.scheduler_add(self.get_shortname(), self.poll_device, cycle=self.cycle, next=shtime.now())
+        else:
+            # use new code from klab for firmware >= 3.x
+            self.scheduler_add(self.get_shortname(), self.poll_deviceV3, cycle=self.cycle, next=shtime.now())
```
bmxp, line 85: Maybe rename the firmware 2 method self.poll_device to self.poll_deviceV2 and assign self.poll_device the correct function?

```
if self.fw == 2:
    # use 'old' code for firmware <= 2.x
    self.poll_device = self.poll_deviceV2
else:
    self.poll_device = self.poll_deviceV3

self.scheduler_add(self.get_shortname(), self.poll_device, cycle=self.cycle, next=shtime.now())
```

solarlog/plugin.yaml
```diff
@@ -3,16 +3,16 @@ plugin:
     # Global plugin attributes
     type: interface                 # plugin type (gateway, interface, protocol, system, web)
     description:
-        de: 'Auslesen der Web-Seite eines SolarLog mit Firmware <= 2'
-        en: 'Read content of Solarlog webpage with firmware <= 2'
+        de: 'Auslesen der Web-Seite eines SolarLog'
+        en: 'Read content of Solarlog webpage'
     maintainer: bmxp
```
bmxp, line 8: maintainer could now be **cipcm** :-)

cipcm: nice try ;-)

solarlog/plugin.yaml
cipcm: I ran the check and removed the mandatory because it's not allowed in connection with default. Also, I renamed the attribute to fw2x which is probably not used anywhere else.

solarlog/__init__.py
cipcm: OK

PR #439, opened 2021-01-03
Update __init__.py

Bonze255:
add self._data['state'] = 'disconnected', when there is no connection to the robot

### DIFF
--- a/xiaomi_vac/__init__.py
+++ b/xiaomi_vac/__init__.py
@@ -78,6 +78,7 @@ class Robvac(SmartPlugin):
     # --------------------------------------------------------------------------
     def _connect(self):
         if self._connected is False:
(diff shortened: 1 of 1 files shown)


msinn:
Shouldn't this be set everywhere where
```
self._connected = False
```
is set and not in the beginning of ``self._connect()``?

Morg42:
Maybe edit the title to refer to the changed plugin, e.g. "Plugin xiaomi_vac: edited ..."

msinn:
The functionality is now implemented in the way I suggested above:
> set self._data['state'] to 'disconnected' everywhere where self._connected is set to False

in commit 98f882442d859a49c2752d722863c8157858c294

This can be closed.

PR #442, opened 2021-01-04
add special datapoint 5999 for Tebis TS systems

wvhn:
Tebis TS was a simple KNX-like bus system sold in the beginning of the 2000s. The special data point 5999 is needed to cover bool and num items.


msinn:
This PR missed some metadata and hat conflicts with the cuurent version of the plugin.
The Functionality is now fully implemented in commit ae0301a7506239c059588b62aa5eba79da9c6616
This PR can be closed.

PR #457, opened 2021-01-15
Stateengine Plugin: version 1.8.0

onkelandy:
Release Notes v1.8.0:
suspend_state in struct: improved duration setting, works with smartvisu countdown widget now
instant_leaveaction can now be overwritten by item attribute se_instant_leaveaction for state items individually
run logic with value action now works ;)
improve queue handling to sequentially evaluate states and run actions without interrupting current evaluations
Major improvement for se_use: can now be defined by lists, values from items and system-wide structs!
se_name is now being updated before checking the state conditions
some improvements to the web interface visualization


msinn:
@onkelandy This is quite a large PR. Shall it be included in the 1.8 release?

onkelandy:
I think it's ready for the release. Will include the previous version nevertheless.

PR #458, opened 2021-01-15
several plugins: updates concerning lib.network

Morg42:
be aware: these changes need the current version of lib.network as provided in smarthome, PR #382!

- modified all plugins using lib.connection to use the (new) lib.network
- updated plugins already using lib.network
- some fixes on the go, code cleanup, documentation

remark: I tried to "save" the old version of some plugins; due to parallel development not all "old" versions are current in relation to release 1.8 This might need to be fixed later if the old versions are to be carried on.

all plugins have been tested by me, @onkelandy or others


Morg42:
Oh, and all plugins using lib.network were modified to enable "proper" restartability as discussen on gitter

PR #468, opened 2021-01-25
[plugins/rcswitch]: BUGFIX in plugin.yaml for shng v1.8.1

hasenradball:
# Bugfix in plugin.yaml
- changed valid list for rc_device to:
   valid_list: ['1','2','3','4','5','a','b','c','d','e','A','B','C','D','E']


dasax:
This change seems to be neccessary for compatibility with shng 1.8. I didn't check so far, as I do not have updated to 1.8 so far. 
I reccomend to accept the change and add hasenradball as maintainer of the plugin.

PR #474, opened 2021-02-05
Avoid exceptions when lastconditionset/lastconditionset_name item is not set

mptei:
If the items lastconditionset/lastconditionset_name are not set and one tries to generate the graph in the web interface an exception occurs.
The pull request avoids the exception. The graph is generated.


onkelandy:
sounds good for me. please always prefix all commits and PRs with "pluginname plugin:"
Thx, will merge after successful test

PR #475, opened 2021-02-06
hue2: add port

mptei:
Adds handling of port number.


bmxp:
I would suggest you put a PR against develop and not against master, please. And consider to increase version number in plugin.yaml and __init__.py as well

PR #476, opened 2021-02-07
Hue2: add port to ip

mptei:
Adds the port number to the ip if it is different from default (80).
Increases version number to be 2.0.0a.


mptei:
Wrong target again ...

PR #477, opened 2021-02-07
hue2: added port to ip; updated version number

mptei:
Adds the port number to the ip if it is different from default (80).
Increases version number to be 2.0.0a.

### DIFF
--- a/hue2/__init__.py
+++ b/hue2/__init__.py
@@ -49,7 +49,7 @@ class Hue2(SmartPlugin):
     the update functions for the items
     """
 
-    PLUGIN_VERSION = '2.0.0'    # (must match the version specified in plugin.yaml)
+    PLUGIN_VERSION = '2.0.0a'    # (must match the version specified in plugin.yaml)
 
     hue_group_action_values          = ['on', 'bri', 'hue', 'sat', 'ct', 'xy', 'colormode']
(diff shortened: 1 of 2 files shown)


mptei:
Just saw at restart:
`plugin 'hue2': Found invalid value '127.0.0.1:88' for parameter 'bridge_ip' (type ip) in /etc/plugin.yaml, using default value '0.0.0.0' instead`

Needs a fix too ...

msinn:
A port number should not be added to an ip. An ip is an ip and nothing else. Like in other Plugins, if a port is needed it has to be implemented as a separate parameter, like other plugins do it.

Since I have a modification to make to the hue2 plugin anyhow, I could add that parameter while I am on it.

That would change the handling of discoverd_bridges as well. So I would incorporate your change in my upcoming modifications.

mptei:
That's OK.

Can hue2 use multiple bridges like the older hue plugin? Then it must be kept in mind that the port configuration is on a per bridge base.

msinn:
As described in the documentation, one instance supports one bridge, but you can configure as many instances of the plugin as you want.

msinn:
Out of curiosity: Why is your bridge running on a different port?

mptei:
This is a deconz  bridge [(dresden-elektronik/deconz-rest-plugin)](https://github.com/dresden-elektronik/deconz-rest-plugin) which is running on a computer together with several other services. Port 80 was already occupied (by nginx).

PR #484, opened 2021-02-21
several plugins: updates concerning lib.network

Morg42:
be aware: these changes need the current version of lib.network as provided in smarthome, PR #397!

- modified all plugins using lib.connection to use the (new) lib.network
- updated plugins already using lib.network
- some fixes on the go, code cleanup, documentation
- remark: some old plugin versions are saved; this might need to be fixed by policy decision

all plugins have been tested by me, @onkelandy or others


bmxp (commented):
Just a small question: Did you change all affected plugins minimum version to the current release? Since lib.network in this flavourthe plugins will not work e.g. with a version 1.8 or lower ...

msinn:
The minimum shng Version for those Plugins should be set to 1.8.2a for now. That signals, that those plugin would not run under shng v1.8.2.

Morg42:
Fixed sh_minversion to 1.8.2a for all edited plugins.

TODO: check if other plugins imported lib.network since 1.8

PR #488, opened 2021-02-22
knx plugin: add DPT 251.600 RGBW

GanfMR:
Implemented & Tested for MDT DaliControl Gateway
6 Bytes Object, but only the 4 Bytes are actually used according to docu. 
The first 2 Bytes are filled with '0xff' as default values

### DIFF
--- a/hue/__init__.py
+++ b/hue/__init__.py
@@ -478,10 +478,10 @@ class HUE(SmartPlugin):
                 # lampe ist an (status in sh). dann können alle befehle gesendet werden
                 if hueSendGroup == 'on':
                     # wenn der status in sh true ist, aber mit dem befehl on, dann muss die lampe auf der hue seite erst eingeschaltet werden
-                    if hueIndex + '.bri' in self._sendLampItems:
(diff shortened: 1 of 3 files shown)


GanfMR:
Oh my second commit was added to this as well, even though I comitted it after my pull request. 
Hue Plugin: BugFix for Groups:
Copy Past Bug Fix. 
Current version does not work correctly for groups, as bri is not set when switching on light.

bmxp (changes requested):
There should be only one commit per plugin and it is easier for the records if you prepend your commit with ``knx plugin: add DPT 251.600 RGBW``

I'd also recommend to bump up the version numbers in ``plugin.yaml`` and ``__init__.py`` as well.

GanfMR:
If I understand you correctly:
- I will close this Pull Request
- I will Rollback my Fork
- Have a new commit, to my Fork as proposed
- Create a new Pull Request?

bmxp:
Yes, this might indeed work. If you think it's too much work I can try to pick your enhancements and merge them into develop.

GanfMR:
Will do so

PR #489, opened 2021-02-24
knx plugin: add DPT 251.600 RGBW

GanfMR:
Implemented & Tested for MDT DaliControl Gateway
6 Bytes Object, but only the 4 Bytes are actually used according to docu.
The first 2 Bytes are filled with '0xff' as default values

### DIFF
--- a/knx/__init__.py
+++ b/knx/__init__.py
@@ -82,7 +82,7 @@ KNXMC = 'IP Router'
 # old class KNX(lib.connection.Client,SmartPlugin):
 class KNX(SmartPlugin):
 
-    PLUGIN_VERSION = "1.7.7"
+    PLUGIN_VERSION = "1.7.8"
 
     # tags actually used by the plugin are shown here
     # can be used later for backend item editing purposes, to check valid item attributes
--- a/knx/dpts.py
+++ b/knx/dpts.py
@@ -448,13 +448,13 @@ Datapoint Types U8U8U8U8
 """
 
 def en251(value):
-    return [0, 0xff, 0x0f, int(value[0]) & 0xff, int(value[1]) & 0xff, int(value[2]) & 0xff, int(value[3]) & 0xff]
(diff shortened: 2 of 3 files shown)


GanfMR:
Sorry, was busy at home & work:
I avoid confusion on the order the Bytes, here how it is also send via ETS:
Byte6 Byte5 Byte4 Byte3 Byte2 Byte1
Byte6: Unused, can be any value
Byte5: Control Byte, where the last 4 Bits indicate if the values in RGBW are valid
Byte4: Red
Byte3: Green
Byte2: Blue
Byte1: White

So Byte5 can be 0x0f, or as used 0xff. 
We could think of exposing Byte5 via KNXD, as the 4 bits can be used to seperatly change color values seperatly without knowing the other values. On the other hand it would make usage in item.yaml more complicated, as everyone has to be aware of the indicator bits.

GanfMR:
There is still a Bug in de251. Did not see that before, due to my item setup. I will fix this. Before I do a new commit, I wait for your view on my last comment

bmxp:
Maybe we should discuss this further on [gitter.im](https://gitter.im/smarthomeNG/smarthome)?

According to the official specifications it is set as follows:
```
Byte 6 (MSB) -> R
Byte 5       -> G
Byte 4       -> B
Byte 3       -> W
Byte 2       -> reserved
Byte 1 (LSB) -> upper 4 Bits reserved, lower 4 Bits indicate if R, G, B, W are valid, separated from each channel.
```

Since I do not have any RGBW devices and no actors in reach to check it you will need to test if it works. I can push this to develop anyway and if we need an adaption we can do so later.
The question would be if one has the need to only set information when the channel bits are 1. So if a telegram is received from knxd where e.g. only W is set to 1, then should SHNG accept only the W channel into the list?

GanfMR:
In my case I listen to telegrams fom the color status, there all channels are send as valid. In cases others do not have that status it could be an issue.
Is it OK to have this 2nd commit, or would you prefer to have a new pull request, even though this conversation would be lost?

Btw: this is out of MDT docu:
grafik

bmxp:
It seems as If we can not geht a final decision here. So I have merged for how.

PR #490, opened 2021-02-25
stateengine: Use path instead of name

mptei:
When an item which is referenced in a action has a name, then an error message appears that the item cannot be found. In the error message appears the name of the item:
`2021-02-24  07:23:45 WARNING  plugins.stateengine.stateengine.automatik.rules      Item 'Here is the name of the item' not found!`

The PR uses the path instead of the name to later find the item ...


mptei:
Sorry. Wrong commit.

PR #491, opened 2021-02-25
Wihtings health: callback base url

mptei:
My smarthomeng is not directly reachable from the internet. Therefore the callback url for authentication is not reachable and authentication is not possible.
Instead I use a proxy to forward the url for the callback to smarthomeng.

This PR adds a new parameter, which allows to specify the basis url for the callback. With this change authentication is possible.
Perhaps it would be even better to specify the full callback url?


psilo909:
Hi, the callback URL can be an internal (LAN) IP/url. the only thing you have to do is to open it in a browser. There is no request from withings towards your SHNG instance

mptei:
Sorry, you have right. I misunderstand that. I thought this callback is checked from outside.

psilo909:
np, enjoy the plugin (first user other than me i think ;-) )

PR #500, opened 2021-03-28
avm: Handle callmonitor reconnect

mptei:
Nearly every night my SmarthomeNG has a connection problem with the fritz box. In this moment an exception occures in the listen thread and the thread terminates. Of couse the call monitor is not working anymore after that.

This commit fixes that behavior as it sets the _listen_active flag to false before terminating the thread. The inherent reconnect behavior then detects this situation and start a new listen thread.

On shutting down SmarthomeNG it stops the listen thread by a rude shutdown of the socket. There is no other way to get in contact with this thread. Alternatively one could set a socket timeout and check for the stop flag ...

Also it now stores a handle to the running listen thread (`start()`still returns None) and uses this for joining.

### DIFF
--- a/avm/__init__.py
+++ b/avm/__init__.py
@@ -82,10 +82,12 @@ class MonitoringService:
             return
 
         self.conn = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.conn.settimeout(None)
         try:
             self.conn.connect((self._host, self._port))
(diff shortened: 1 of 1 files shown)


psilo909:
I will check you changes at latest over the easter weekend!

mptei:
An easy test is to switch off the call monitor and later on again ...

psilo909:
testing the plugin over the weekend. just integrated the code and checked it manually. so far all fine, thx for the update! will merge in case of no problems on sunday or monday

PR #502, opened 2021-04-05
PR: BSB-LAN

thengsty:
- Compatibility check for new BSB-LAN Version 2.x.
- Minor tweaks on README file


bmxp:
Would be great to have a user_doc.rst in future (german)...

PR #504, opened 2021-04-14
Functional Update of snmp plugin incl enhancement of WebIF

sisamiwe:
Hi, 

this a a PR for a functional update of the SNMP plugin mainly adding support for printers.
Furthermore the WebIF has also been reworked

Feedback is appreciated.


snmp/__init__.py
```diff
@@ -60,7 +65,7 @@ def __init__(self, sh, *args, **kwargs):
         self.instance = self.get_parameter_value('instance')         # the instance of the plugin for questioning multiple smartmeter
         self.cycle = self.get_parameter_value('cycle')               # the frequency in seconds how often the query shoud be done
         self.host = self.get_parameter_value('snmp_host')            # IP Adress of the network device to be queried
-        self.port = self.get_parameter_value('snmp_port')            # Port for SNMP queries
+        self.port = int(self.get_parameter_value('snmp_port'))       # Port for SNMP queries
```
bmxp, line 68: Why do you need to convert to int? Could you possibly use int instead of num in plugin.yaml?

sisamiwe:
@bmxp 
THX for reviewing the PR.
I've implemented the proposed change. Now all checks have passed.

Michael

PR #534, opened 2021-10-27
rpi1wire Plugin:

sisamiwe:
- Updated User Docu
 - Updated WebIF
 - COde Cleanup


msinn:
I found some errors in the documentation of the plugin:

- the documentation file must be called **user_doc.rst** not user_docu.rst
- readme.rst ist not a supported file (there used to be the need for README.md, but that was long ago)
- referenced jpegs are missing (should be in **user_doc/assets/** directory) Found pictures with the right filenames in **webif/static/img/**, which are not referenced anywhere

PR #542, opened 2021-12-12
UZSU extensions for series

Andrek01:


onkelandy (changes requested):
There are some issues. The imports need to be there as they are. Additionally, the lastvalue handling is reverted in this PR - please revert to the latest version.
Additionally, I'd put the plugin version writing in the run method.

PR #571, opened 2022-02-03
AVM-Plugin update:

sisamiwe:
Update Docu
Update WebIF
Connect to Smarthome Devices via AHA-Interface (tested mit Fritz 440, Comet Dect) (therefore Plugin AVM smarthome can be depreciated)
implement call deflections (tested mit Fritzbox 7530)
Rename Parameter "index" to "avm_tam_index"
Code Cleanup (add new functions to mimimize code repetitions)
Improve error handling


sisamiwe:
So here is an new try of a PR. 
The checks failed, but I don't know why. Is there anybody who can help?

sisamiwe:
I've done another one with passed check and better tracebility. Now it should work.

PR #575, opened 2022-02-08
Uzsu Plugin 1.6.1

onkelandy:


uzsu/__init__.py
```diff
@@ -536,12 +552,14 @@ def _schedule(self, item, caller=None):
                               " Use the latest device.uzsu from SV 2.9. "
```
bmxp, line 552: SV 2.9 or above?

bmxp (approved):
Seems to be fine though I have no chance to test it currently ...

uzsu/__init__.py
onkelandy: done

PR #587, opened 2022-02-20
Super smooth dpt3 dimming for deconz (hue2)

CaeruleusAqua:
Hello together,

today I have reconfigured all my lights in deconz, so that they are off after an power cycle (in our house there was a power outage last night and somehow it shitty that in the middle of the night almost all the lights in the house turned on). While doing that I noticed that Zigbee's "level control cluster" directly supports a relative dimming (with dimmstart and stop). After reading the code of deconz's REST API for a while, I found out that the function is implemented via the bri_inc attribute (0 means stop dimming).

I had to extend the HUE 2 plugin a bit (add bri_inc and hue2_transitionTime).

Furthermore I added the possibility to create a reference light for groups. If this is specified, the values are read from the reference light and not from the group action.
This was necessary because the values in the group action are not updated after the relative change.

The procedure for dimming corresponds in principle to the dimming in the old hue plugin:

Example:

```
KITCHEN:
    name: spots
    hue2_id: 7
    hue2_resource: group
    level_inc:
        type: num
        hue2_resource: ..:.
        hue2_id: ..:.
        hue2_function: bri_inc
        hue2_transitionTime: 6
        enforce_updates: 'true'
        dimdpt:
            type: list
            knx_dpt: 3
            knx_listen: 1/1/11
            dpt3_dim: true
            enforce_updates: 'true'
    level:
        type: num
        hue2_resource: ..:.
        hue2_id: ..:.
        hue2_function: bri
        hue2_transitionTime: 0.2
        hue2_refence_light_id: 3
```
          
I use Deconz, whether the whole thing also runs with the original HUE bridge I can not evaluate.
Maybe someone want to test this?

Limitations:
- Relative dimming range goes with me only from 1 to 254
- level/bri is set to 1 (dim down) or 254 (dim up) until the end of the dimming process, but then updated to the correct value.

### DIFF
--- a/hue2/__init__.py
+++ b/hue2/__init__.py
@@ -49,12 +49,12 @@ class Hue2(SmartPlugin):
     the update functions for the items
     """
 
-    PLUGIN_VERSION = '2.1.0'    # (must match the version specified in plugin.yaml)
+    PLUGIN_VERSION = '2.2.0'    # (must match the version specified in plugin.yaml)
 
-    hue_group_action_values          = ['on', 'bri', 'hue', 'sat', 'ct', 'xy', 'colormode','alert', 'effect']
-    hue_light_action_writable_values = ['on', 'bri', 'hue', 'sat', 'ct', 'xy']
-    hue_light_state_values           = ['on', 'bri', 'hue', 'sat', 'ct', 'xy', 'colormode', 'reachable','alert','effect']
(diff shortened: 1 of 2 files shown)


bmxp:
@CaeruleusAqua Sorry but I forgot to mention that it is needed to do PR agains develop and not against master. We keep master clean for releases only. Could you please repeat your PR against develop branch?
Maybe @msinn can have a look at your PR then

CaeruleusAqua:
Done!

PR #590, opened 2022-03-13
Super smooth dpt3 dimming for deconz (hue2)

CaeruleusAqua:
Hello together,

today I have reconfigured all my lights in deconz, so that they are off after an power cycle (in our house there was a power outage last night and somehow it shitty that in the middle of the night almost all the lights in the house turned on). While doing that I noticed that Zigbee's "level control cluster" directly supports a relative dimming (with dimmstart and stop). After reading the code of deconz's REST API for a while, I found out that the function is implemented via the bri_inc attribute (0 means stop dimming).

I had to extend the HUE 2 plugin a bit (add bri_inc and hue2_transitionTime).

Furthermore I added the possibility to create a reference light for groups. If this is specified, the values are read from the reference light and not from the group action.
This was necessary because the values in the group action are not updated after the relative change.

The procedure for dimming corresponds in principle to the dimming in the old hue plugin:

Example:

```
KITCHEN:
    name: spots
    hue2_id: 7
    hue2_resource: group
    level_inc:
        type: num
        hue2_resource: ..:.
        hue2_id: ..:.
        hue2_function: bri_inc
        hue2_transitionTime: 6
        enforce_updates: 'true'
        dimdpt:
            type: list
            knx_dpt: 3
            knx_listen: 1/1/11
            dpt3_dim: true
            enforce_updates: 'true'
    level:
        type: num
        hue2_resource: ..:.
        hue2_id: ..:.
        hue2_function: bri
        hue2_transitionTime: 0.2
        hue2_refence_light_id: 3
```
          
I use Deconz, whether the whole thing also runs with the original HUE bridge I can not evaluate.
Maybe someone want to test this?

Limitations:
- Relative dimming range goes with me only from 1 to 254
- level/bri is set to 1 (dim down) or 254 (dim up) until the end of the dimming process, but then updated to the correct value.


msinn:
Do you have an example how to configure/use the ``function: dict`` feature, you have implemented?

CaeruleusAqua:
Sorry, the function "hue - Multiple functions at once" should not yet be included in the pull request. I will create a new one with a bugfix, which also contains some documentaion.

CaeruleusAqua:
I just saw that there is a reference to this pull reqeuest in the documentation.

The Dict function allows multiple parameters to be sent to the Hue Api on a single time.
So setting brightness or color temperature at the same time instead of sequentially, which at least for me (deconz) leads to much smoother transitions.

In addition, it is also possible to set the transition time individually for each call:

e.g.: 
{'on': True, 'transitiontime': 10}

All tuples in dict are passed one-to-one to the qhue library.

So at least (but not exclusively) the following values should be able to be in the dict:
on, bri, hue,sat,ct,transitiontime

PR #591, opened 2022-03-13
smlx: convert EMH ED300L values to nicer format

julianze:
Hello together,

i was migrating from smarthomeng v1.5.1 to v1.9.1 and now using the smlx plugin. Before I have used the sml plugin.
I noticed WARNINGs about an incompatible string (manufactor + serial number) coming from the smlx plugin to the item.

`
2022-03-13 20:10:45 CET WARNING  item              plugins.smlx.Smlx Item Energie.Strom.Zaehler.Hersteller: value "b'EMH'" does not match type str. Via Sml None  --  (item.py:__update:1536)  
`

This pull request will fix this by expanding the obis codes for EMH ED300L smartmeter to convert the byte values to a nicer format.

This is tested on my installation and it works quite fine.

Happy to see this in a future release.

Kind Regards,
Julian

### DIFF
--- a/smlx/__init__.py
+++ b/smlx/__init__.py
@@ -419,11 +419,12 @@ class Smlx(SmartPlugin):
                     # For a Holley DTZ541 with faulty Firmware remove the                ^[1] from this line ^.
 
                     # Convert some special OBIS values into nicer format
+                    # EMH ED300L: add additional OBIS codes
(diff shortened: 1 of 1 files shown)


bmxp:
To help out with this I merge this PR. However to have hard coded conversions should be avoided. Maybe there is a more general way to support conversions.

PR #593, opened 2022-03-20
Sonos plugin: Subprocess ping not working

RedTiger26:
The Sonos plugin stopped working for me. Debugging the issue, I found that the subprocess used for ping does return a return code 1, despite the fact the speakers where online and pingable from the command line.

I switch the command used in the subprocess module to run, which is recommended since Python 3.5 and got it working again that way. To help debugging these kind of issues, I also added two debugging log messages in this area.


aschwith:
PR looks ok. Can you please clarify the reasons that caused the sonos plugin to stop working? Did you upgrade python and if yes, from to which version?

aschwith:
Basic contributeion has been integrated with commit "3c6d7ef2710c4dbe52224b06e7abdb94cff81bad". Closing this one. Thanks for the contribution.

PR #601, opened 2022-04-19
KNX dpts: RAW

laraveluser:
This function offers the possibility to read RAW data from the KNX. This is necessary, for example, to be able to read out key IDs from (old) Winkhaus locking systems.

### DIFF
--- a/knx/dpts.py
+++ b/knx/dpts.py
@@ -32,6 +32,15 @@ from ``The KNX Standard v2.1`` chapter ``03 Volume 3 System Specifications``.
 """
 
 """
+Datapoint type raw
+"""
+
+def deraw(payload):
+    import binascii
+    return binascii.hexlify(payload).decode()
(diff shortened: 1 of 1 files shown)


bmxp:
Thank you for your PR!
Unfortunately you've put it against master branch. Could you please repeat against develop branch?
We use master branch only for releases.

And if I may ask to increase the version of the plugin in ``__init__.py`` and ``plugin.yaml`` as well?
And furthermore add the dpt to plugin.yaml, too?

msinn:
The other dpts also allow sending them (by implementing an encode method. Wouldn't it be sensible to do it for the **raw** dpt too?

msinn:
Reimplemented this PR with commit https://github.com/smarthomeNG/plugins/commit/5bd11dee9fe2d7c97ef903783446a32543c82fc0

PR #603, opened 2022-04-25
Added new husky plugin wich uses actual husqvarna api

gruberth:


husky2/plugin.yaml
```diff
@@ -0,0 +1,252 @@
+# Metadata for the plugin
+plugin:
+    # Global plugin attributes
+    type: interface                   # plugin type (gateway, interface, protocol, system, web)
+    description:
+        de: 'Husqvarna Automower Plugin'
+        en: 'Husqvarna automower plugin'
+    maintainer: <autor>
```
bmxp, line 8: maybe you can add here  a support thread, a link to the documentation for user_doc and a maintainer?

bmxp (commented):
Did you make the pictures that you have provided? I fear that we could get some claims from Husqvarna if we simply use them and they are not taken by yourself.
At least a copyright and usage notice seems to be necessary. Maybe you get permission from them and include it in the docs if so?

husky2/plugin.yaml
gruberth: No, I used only the pictures wich where already in the old husky plugin. Would it be enough if I make a watermark or something on them?

gruberth:
No, I used only the pictures wich where already in the old husky plugin. Would it be enough if I make a watermark or something on them?

msinn:
Since the pictures are already in the old plugin, no additional harm is taken by accepting the PR. The permission should be asked for though. If the permission is not given by huskvarna, the pictures should be replaced by clipart w/o copyright in the future. 

Usually the plugins use clipart from openclipart.org or similar sites.

PR #607, opened 2022-05-02
husky2: Final updates

gruberth:
I removed the husqvarna images becase I am not getting any answer of husqvarna. These images where only in the webif to give a visual hint to the used model. But I think everybody knows his mower, so an additional picture is unneccessary. If I get the permission someday I will put them back into the plugin.

Since I am now into the code, I can also remove the images form the origingal husky plugin and make a PR, if that is desired.


msinn:
You wrote that the old plugin isn't working any more. If that is the case, we should mark the old plugin as deprecated in the upcoming maintenance release and remove it with the next feature release of SmartHomeNG. 

If we deprecate the old plugin,  it is not necessary to remove the pictures from the old plugin.

Is it the case, that the old plugin doesn't work any more and is of no future use?

gruberth:
Yeah this is correct, the plugin does not work anymore. The old api, on wich the original husky plugin is based, has been shut off.

Thank you @msinn for your additional work and fixes on my PR.

PR #617, opened 2022-06-18
Update requirements.txt

henfri:
pyotp is a missing requirement


bmxp:
could you please create a PR against develop, not against master?

PR #624, opened 2022-09-03
stateengine plugin: adapt widget to new SV icon handling

wvhn:


wvhn:
will be replaced with changed commit message

PR #630, opened 2022-09-23
plugin SV widgets: set proper defaults in stateengine widget / update widget imports

wvhn:
Twig / php otherwise throws warnings


wvhn:
stateengine plugin: set proper defaults in SV widget
rtr plugin: update SV widget import
rtr2 plugin: update SV widget import
simulation plugin: update SV widget import
openweathermap plugin: update SV widget import

PR #643, opened 2022-10-28
SHNG min 1.8

CannonRS:
- Minium SmartHomeNG 1.8
- changes.txt not longer used and removed


msinn:
Changes from this PR are already included in plugin. (Have been added by a concurrent push)

PR #646, opened 2022-11-06
Update for compatibility to pymodbus version >= 3.0.0

beckerth:
Pymodbus:
Version [2.5.3](https://github.com/riptideio/pymodbus/releases/tag/v2.5.3) is the last 2.x release (Supports python 2.7.x - 3.7).

Version [3.0.2](https://github.com/riptideio/pymodbus/releases/tag/v3.0.2) is the current release (Supports Python >=3.8).

### DIFF
--- a/kostalmodbus/README.md
+++ b/kostalmodbus/README.md
@@ -27,7 +27,7 @@ This plugin connects your Kostal inverter (https://www.kostal-solar-electric.com
 | PLENTICORE plus 5.5 | yes          | no      |
 | PLENTICORE plus 7.0 | yes          | yes     |
 | PLENTICORE plus 8.5 | yes          | yes     |
-| PLENTICORE plus 10  | yes          | no      |
+| PLENTICORE plus 10  | yes          | yes      |
 | PIKO IQ 4.2         | yes          | no      |
 | PIKO IQ 5.5         | yes          | no      |
 | PIKO IQ 7.0         | yes          | no      |
--- a/kostalmodbus/inverter.py
+++ b/kostalmodbus/inverter.py
@@ -7,7 +7,7 @@ import logging
 import pymodbus
 
 from lib.model.smartplugin import *
-from pymodbus.client.sync import ModbusTcpClient
+from pymodbus.client import ModbusTcpClient
 from pymodbus.payload import BinaryPayloadDecoder
 from pymodbus.constants import Endian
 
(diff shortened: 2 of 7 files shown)


msinn:
As I understand it, pymodbus >= 3.0 is only supported unter Python 3.8 and up. Am I right?
In this case you should add the minimum Python version to the Plugin's metadata. Otherwise people still using Python 3.7 could run into error, which they possibly don't understand.

Tom-Bom-badil:
Are we now also taking over the tasks of 3rd party developers and monitor / update any possible future combination of python and pymodbus in our requirements?

`pip3 install pymodbus` will automatically choose and install the right pymodbus version for the python version found in place.

So, technically all the plugins that have adopted @onkelandy's code suggestion will run fine with *any* version of pymodbus 2.x onwards, no matter which python version is in use.

That is also the reason why I reverted back all that 'pymodbus.bla >= python.blubb' to a simple, single line in my requirements.txt: `pymodbus >= 2.2.0`:

- if you are below pymodbus 2.2 or have no pymodbus at all, the requirements check will fail,
- if requirements check is ok, it will try load pymodbus2, and if it fails, it will load pymodbus3.

That simple.

/tom

msinn:
> Are we now also taking over the tasks of 3rd party developers and monitor / update any possible future combination of python and pymodbus in our requirements?

We are doing that for a quite long time.

The problem with the requirement of this PR is, that it cuts out all users using Python 3.7, since no pymodbus is installed for Python versions before 3.8. This will throw an exception.

Btw.: PR #658 implements support for pymodbus 2 and 3 in the plugins in question.

Tom-Bom-badil:
> Btw.: PR #658 implements support for pymodbus 2 and 3 in the plugins in question.
*gasp* You missed my point.

This plugin **only** implements the new beaviour of pymodbus 3.x.

`result = self.client.read_holding_registers(adr_dec, 2, _unit_=71) #pymb 2.x`
is hardcoded changed to
`result = self.client.read_holding_registers(adr_dec, 2, _slave_=71) #pymb 3.x`
without any further evaluation.

This will only work with pymodbus 3.x and NOT with the former 2.x. So, for the code uploaded in this PR, the statement `pymodbus >= 3.0.0` is technically correct. 2.x is not supported anymore - not during loading, nor during operation/polling.

If the plugin developer is willing to invest the time for dual support 2.x/3.x (which I think is up to the developer to decide upon), he could simply revert back to `pymodbus >= 2.3.0` as it was before, while PR #658 as mentioned by you suggests the following:

```
pymodbus==2.5.3;python_version<"3.8"
pymodbus>=3.0.2;python_version>="3.8"
```

This more complex code threw errors here during testing of my revised Modbus plugins with different Python versions. So I reverted back to the simple solution for my own PR #662. 

Anyways, not my cup of tea, will stop commenting at this point.

/tom

beckerth:
Closed in favor of PR https://github.com/smarthomeNG/plugins/pull/658.

PR #661, opened 2022-11-27
Helios Plugin: Included additional attributes in plugin.yaml

Tom-Bom-badil:
This change will not affect any functionality at all and is just a 'convenience patch'.
The 'new' attributes are defined within the plugin at runtime and have already been around since >8 years.
At some point, they started to spam the logfiles with 'attribute undefined' errors for each single one of them.
Consequently, they are now included in the plugin's attribute list to avoid those unnecessary logfile entries.


Tom-Bom-badil:
Closed, new combined PR coming.

PR #669, opened 2022-12-18
[enocean]: typo correction in eep_parser

hasenradball:
# Changes
- correction of some typo
- added some logger


aschwith:
This PR includes multiple changes of different plugins. Please do one PR per plugin.

hasenradball:
I think you didn`t had a look to the files which was changed!
If you had a look there you would have seen that only the `eep_parser.py` was changed.

aschwith:
Either you come up with a new clean pull request or things won't get merged. I can live without this contribution. It's as simple as that.
If you decide to do a new PR, please leave the plugin log levels as they are and only stick to the rest of your typo corrections.

hasenradball:
@aschwith 
What is now the problem of the PR?
Is it the log-level or is it `This PR includes multiple changes of different plugins. Please do one PR per plugin`?
Do you know what you want?

hasenradball:
@aschwith  
I would give you an small hint for your enocean plugin.
If you take a look at the init.py file it looks like a mess.
Mostly all new methods are added to the init file, also the Webif interface.
Others do it in a separate package and import it when needed.

I would recommend to think about to rework the enocean plugin completely so that the init file becomes clean again.
Therefore you need a structure in mind, how should it look like.
Also the constants at the beginning of the file => place them in a separate package and import it by one single line.
Then the main file becomes cleaner and cleaner.

And one hint, sometimes it is real good adding some multiple line comments, to describe in a nutshell what the method is designed for, etc...
 [documenting python](https://realpython.com/documenting-python-code/)

Have a look to other enocean python implementations.
The base file is clean, the necessary methods are imported from the relevant src files (communicators, send methods, read methods, etc...)

If you add all the time all new methods to the init file it becomes worse and worse by the time.

And if you are willing to do so, maybee you need support by other contributors if you are open to this.

Best regards
hasenradball


PS: here you find some good example
[good structured enocean python](https://github.com/kipe/enocean)

PR #676, opened 2022-12-28
tasmota plugin

stoepf:
new sensors:
- adc input
- esp32 internal temperature sensor

uptime fix


stoepf:
Another pull request with major version change exists

PR #682, opened 2023-01-26
Develop - enhancement - added autocomplete to Executor-plugin (as in shNG-logics)

Andrek01:


Andrek01:
Did a mistake and started a PR directly to "develop" - so i canceled everything.
For the enhancement for Executor i will start a new PR

PR #701, opened 2023-02-17
AVM Plugin Changes:

sisamiwe:
- complete re-writing of plugin, putting TR064, AHA und CallMonitor in separate Classes for better readability and maintenance
- Available TR-064 services and actions the the AVM device will be read at startup from device itself and are in general available within the plugin
- all known/current item attributes and functions are available
- deprecated warning for some item attributes will be put to logger; please change to new attributes (e.g. ain --> avm_ain)
- items with avm_attribute causing errors (e.g. read of deflection of not set) will be blacklisted after error occurred 3 times, Reset of blacklist is available during runtime via WebIF
- some new functions implemented - avm_read_after_write: read the value back after writing - avm_data_cycle: use if item shall have different poll cycle the standard; definition is in seconds; 0 - means just read after init - some more

The content of that PR has already been intensivly testet by @onkelandy and @psilo909 .
@aschwith please test that PR as well as discussed.

@msinn: having that PR released, the plugin "avm_smarthome" can be declared as deprecated.

THX


psilo909:
Über mehrere Wochen getestet und aus meiner Sicht fehlerbereinigt. Funktioniert redundant zur alten Version bei mir.

PR #707, opened 2023-03-05
add new sml2 plugin with asyncio, based on libsml

CaeruleusAqua:
Hello together,

Actually I only wanted to add the asyncio in the smlx plugin, because serial frames were regularly broken by polling.

But then I noticed that the whole SML parser has quite a lot of bugs. (among others memory leaks, which didn't cause any problems until now, because all data is discarded after the polling cycle). After it became more and more problems, I saw that there is a SML lib for python and decided to create a new plugin based on libsml. Unfortunately, I could not adapt all the "features" of the smlx plugin with the lib, as the parser has individual adaptations for specific energy counters.

Because the plugin no longer contains a parser, it is much simpler than the "old" smlx plugin.


msinn:
I cannot do any functional checking, since I don't have the hardware. Since it is a new plugin, I see no harm in merging the PR. Maybe @bmxp can take a look at the plugin after tghe PR is merged.

Formatting of the section titles in user_doc.rst isn't in sync with our documentation standards. I am going to correct it after accepting the PR.

A tip for the future: In the tools directory of the develop branch is a new tool, which does some formal checking of a plugin. The tool check_plugin.py checks:
- the metadata (quite complete check)
- the code (only rudimentary right now)
- the documentation user_doc.rst  (only rudimentary right now)

Feel free to take a look at the [documentation](https://smarthomeng.github.io/dev_doc/entwicklung/plugins/plugin_checker.html).

bmxp:
I was in the process of rewriting the whole parser but got stuck two years ago and did not have time to proceed on this. So if this plugin works for you I am glad to test it as soon as I can. But this might take some weeks. Let's just merge it into next master and I review and test as soobn as possible. 
Thanks for sharing!

CaeruleusAqua:
Welcome, Unfortunately I don't have much time either, the code has been lying around for a few months and I've only just got around to creating the pull request.

PR #731, opened 2023-03-30
[alexarc4shng] Version 1.0.3

Andrek01:
Last Update with MFA, tested with smarthomeNG develop today on 17:00


msinn:
This PR contains 3 Plugins: 
- updates for: **alexarc4shng**, **indego4shng**
- new plugin: **mieleathome** (state 'develop')

I'll look into it.

Btw.: Is there a special reason why your SmartHomeNG plugins have '4shng' in the plugin name?

Andrek01:
Hi Martin,
one reason was  in the beginning of the "new" indego-Plugin there was the need to have new name to prevent collissions with the original and there was a problem with item´s named indego... and the plugin-name only indego.
For alxearc i started in the same way. But there was no special reason.

best regards
Andre

PR #740, opened 2023-04-08
[RCS1000N]: Bugfix -> improve TRY and EXCEPT clauses

hasenradball:
- improve try and except clauses to improve find error root cause.


hasenradball:
forgot to sync fork :-(

PR #745, opened 2023-04-13
kodi: replace with kodi sdp version

Morg42:
Please test functionality and config against your setup - plugin and item config will probably need updating.

Will merge only after successful testing (and possibly debugging)


onkelandy:
In general this works fine. Only issue is the multiinstance functionality. Reading groups don't work with that. 

Another issue when getting large data back (e.g. settings info) the result is hacked in chunks. This is a network lib issue, however.

Morg42:
Issues should be resolved, multiinstance is working fine here.

Struct-generation will be fixed inside sdp itself

PR #750, opened 2023-04-19
modbus_rtu: New plugin

gruberth:
Added new plugin modbus_rtu.
Adapted from modbus_tcp to communicate over modbus rtu with simple usb adapters.


gruberth:
Going to fix error for python 3.7 first...

bmxp:
@gruberth Would it not be better to merge modbus_tcp and modbus_rtu within a single modbus plugin? If we have two plugins it is needed to care twice for the same code...?

gruberth:
Yes you are right, but for me the even bigger problem at the moment is writing to a slave. I spend a ton of hours trying to get pymodbus to send data correctly to a slave, but I was not able to do this over rtu. Reading data works great. Maybe it's because my device doesn't echo back the data on receive. I switched now over to the umodbus lib, where I don't have any problems at all.

Have you or someone else encountered similar issues on modbus tcp?

bmxp:
Unfortunately I have no device to write to with modbus tcp, just an inverter to query.

PR #755, opened 2023-04-27
telegram V1.8.0 new lib package (V20.2) async

ivan73:
Unfortunately, the telegram updater creates a thread called "asyncio_0", which I haven't managed to give a better name to.


onkelandy:
I've tested this successfully on my Raspi 4. However, I recommend three changes:
- add the old version in a subfolder so it is not "gone forever"
- remove the parent <div> for the datatable in the index.html as this should be included now automatically.
- The ' in ID's should be removed

onkelandy:
Thanks for the update. Could add a correct locale.yaml for the old (1.7.1) and new version please?

bmxp (commented):
I think it is not a good option to evaluate once per plugin start if the debug is enabled or not. 
It is possible to change logging during runtime with admin interface. 
I think it is a better way to evaluate if debugging is anabled once per function but not at startup only.

bmxp:
And I would also think that this plugin should be something like 2.0 instead of 1.8 since the internal structure changed with async io a lot.

Another question: In the old telegram library with threads it somestimes seems I am missing some messages due to network troubles at telegram. Any ideas if the messages can be cached locally until it is proven they have been received by telegram servers?

ivan73:
> And I would also think that this plugin should be something like 2.0 instead of 1.8 since the internal structure changed with async io a lot.
> 
> Another question: In the old telegram library with threads it somestimes seems I am missing some messages due to network troubles at telegram. Any ideas if the messages can be cached locally until it is proven they have been received by telegram servers?

When the Telegram message is successfully sent to the server, a "Message sent:" entry is logged using logger.debug. In case of a failed transmission, a logger.error ""could not broadcast to chat id " entry is made. This allows for identifying if occasionally a message was not sent. In the case of a failed transmission, the message can be written back to the send_queue to trigger a resend. However, this will result in continuously (every second) repeating sending attempts in the case of an unreachable server or an invalid chat_id. Would this be acceptable?

onkelandy:
However, this will result in continuously (every second) repeating sending attempts

Could this be made configurable?

onkelandy:
Braucht es da wirklich die Exception auch? Aus meiner Sicht würde der Error genügen, sagt eh schon alles aus...
```
Mai 15 14:01:53 ERROR    telegram.ext.Updater Error while getting Updates: Bad Gateway
Mai 15 14:01:53 ERROR    telegram.ext.Updater Exception happened while polling for updates.
Traceback (most recent call last):
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/ext/_updater.py", line 607, in _network_loop_retry
    if not await action_cb():
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/ext/_updater.py", line 335, in polling_action_cb
    raise exc
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/ext/_updater.py", line 320, in polling_action_cb
    updates = await self.bot.get_updates(
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/ext/_extbot.py", line 543, in get_updates
    updates = await super().get_updates(
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/_bot.py", line 381, in decorator
    result = await func(self, *args, **kwargs)  # skipcq: PYL-E1102
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/_bot.py", line 3661, in get_updates
    await self._post(
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/_bot.py", line 469, in _post
    return await self._do_post(
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/ext/_extbot.py", line 325, in _do_post
    return await super()._do_post(
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/_bot.py", line 497, in _do_post
    return await request.post(
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/request/_baserequest.py", line 168, in post
    result = await self._request_wrapper(
  File "/home/smarthome/.local/lib/python3.9/site-packages/telegram/request/_baserequest.py", line 332, in _request_wrapper
    raise NetworkError(description or "Bad Gateway")
telegram.error.NetworkError: Bad Gateway
```

onkelandy:
Falls das aktuelle locale.yaml auch für die vorige(n) Version(en) passt, würde ich's auch in die Unterordner kopieren.

ivan73:
> Falls das aktuelle locale.yaml auch für die vorige(n) Version(en) passt, würde ich's auch in die Unterordner kopieren.

das passt nicht, ich habe 4 Übersetzungen im Unterordner welche das Webinterface betreffen hinzugefügt

ivan73:
> Braucht es da wirklich die Exception auch? Aus meiner Sicht würde der Error genügen, sagt eh schon alles aus...

hab es versucht mit error_callback abzufangen - könntest du das bitte testen?

ivan73:
"Some checks were not successful" soll ich den pullrequest löschen und einen neuen starten - wo die ganzen commits zusammengefasst wären?

onkelandy:
Die Fehler haben eine andere Ursache. Lass mal

PR #756, opened 2023-04-29
Appletv Plugin: Updates and fixes

onkelandy:


Foxi352:
Hi. I was also working on an update to newest pyatv and other things like separating webif init.py.
My current code status does detect all my apple devices like Homepods or iPhones, but no Apple TV's.
Does your code work  so far ? If yes we could merge this PR and i will just redo the parts which i did and which are not in your codebase ....

onkelandy:
Currently, user_doc, web if and plugin works the way is is expected.. more or less. I got some issues with sending RC commands, etc. but not sure where they originate from.

However, as a minimum starting point for your adaptations this PR would be fine.

However, there is an issue with the zeroconf requirement in the "all.txt". It seems like hue2 needs a lower version of it than pyatv. Not sure how to proceed here..? @msinn do you know what happens with hue2 with a newer version of zeroconf (on python 3.9+?

Foxi352 (approved):
I suggest, as discussed, that you merge the changes and i will continue based on your work.

onkelandy:
The tests should succeed first. But if you want to work further based my changes you could just download my branch.
I'll be away now until the end of the week.

PR #773, opened 2023-07-02
DB_ADDON Plugin: Code rework to enable more flexible attributes

sisamiwe:
- bump to version 1.2.0
- rework code to create query parameter during parse_item
- harmonize parameters to ease handling of queries
- enable onchange items also to be calculated during start
- bugfixes


onkelandy:
reviewed and tested a lot, looks pretty good now!

PR #793, opened 2023-08-09
Support RFC2217 connections for enocean plugin

JHthe4:
This change is in reference to [this thread](https://knx-user-forum.de/forum/supportforen/smarthome-py/1872963-verlust-von-serial-device-bringt-enocean-und-smarthomeng-zum-absturz) at the KNX user forum. 

The `serial.Serial()` constructor from pyserial was replaced with `serial.serial_for_url()`, which supports both the traditional unix file handle (e.g. `/dev/ttyAMA0`) and RFC2217-compliant URLS (e.g.`rfc2217://<IP>:<PORT>`) interchangeably. 

Error handling was added to deal with connection breakdowns and reinitiate the connection. In case of a connection breakdown at transmission time, one attempt at retransmission will be made. Afterwards, the packet will be lost.

It's not clear to me if the run method will be restarted on successful reconnection after the loop has been broken (change made in commit c9c2538 ), so this might be a potential improvement for the future.

### DIFF
--- a/enocean/__init__.py
+++ b/enocean/__init__.py
@@ -166,7 +166,7 @@ SENT_ENCAPSULATED_RADIO_PACKET = 0xA6
 
 class EnOcean(SmartPlugin):
     ALLOW_MULTIINSTANCE = False
-    PLUGIN_VERSION = "1.3.9"
+    PLUGIN_VERSION = "1.3.10"
 
     
     def __init__(self, sh, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


aschwith:
Thx for the contribution. Very much appreciated. I will test the PR in the next couple of days regarding backwards compatibility. Code changes look ok. Would you mind adding a few lines of explanations to the user_doc, e.g. by adding the IP harware to the compatibility list and extending the section about the serial port to a more generic way that covers your new feature?

JHthe4:
No problem :). I've expanded the documentation to contain an example setup and explain the new config option. Please let me know if there's anything missing.

FYI, we've had this running on our system for about two months now and it's been very resilient thus far. If there's anything we can do to help test, feel free to reach out :).

PR #794, opened 2023-08-09
update vacations plugin

stoepf:
update after start
update cycle configurable
accept country "Germany"

### DIFF
--- a/vacations/__init__.py
+++ b/vacations/__init__.py
@@ -86,7 +86,7 @@ class Vacations(SmartPlugin):
 
     def get_vacation(self, date_str=None, province=None):
         if province is None:
-            if self.shtime.config.get('location', None).get('country') not in ['DE', 'de', 'Germany']:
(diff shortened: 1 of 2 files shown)


vacations/__init__.py
```diff
@@ -85,7 +86,7 @@ def get_province_codes(self):
 
     def get_vacation(self, date_str=None, province=None):
         if province is None:
-            if self.shtime.config.get('location', None).get('country') not in ['DE', 'de']:
+            if self.shtime.config.get('location', None).get('country') not in ['DE', 'de', 'Germany']:
```
psilo909, line 89: would it make sense to check for "germany" and "Germany" here? I am not 100% sure anymore where the value comes from. is it resolved by lat and lon? or custom set? or can it be both?

psilo909: ah i see its from location.yaml and "Germany" is listed as valid one.. in any case i think "germany" makes also sense here

stoepf: country is converted to lowercases now, so it is not case-sensitive

PR #820, opened 2023-09-15
Stateengine Plugin: fixes and improvements

onkelandy:


onkelandy:
Needs to be tested before merging. Last commit needs update

PR #827, opened 2023-09-21
Avm update

sisamiwe:
AVM: Add set_color and set_hsv to enable setting color and color/brightness

- bump to 2.2.0
- update user_doc.rst showing all allowed attributes
- enhance item_attributes_master.py to add/update attributes list and description
- add / enhance methods to set color and hsv
- enable to switch to on/off based on dimmer value


aschwith (approved):
Tested for more than three weeks in productive environment. All fine.

PR #833, opened 2023-09-25
Stateengine Plugin: new se_stateorder functionality, list items, more updates

onkelandy:
needs to be tested by @Morg42


Morg42:
Looks good to me, works here with alle previous functions and state reordering at runtime

PR #844, opened 2023-10-29
knx: Add dpt 5003

CaeruleusAqua:
I am added support for dtp 5003 to the knx plugin, 

as discussed in:
https://knx-user-forum.de/forum/supportforen/smarthome-py/1702849-knx-dpt-5-003


bmxp:
@msinn I have no test case to check but it looks as if it will work as intended. I think it can be merged

PR #851, opened 2023-11-02
Influxdb2: add web interface screen shot and conversio script to docu

onkelandy:


gruberth:
Aaah its already merged...
... But should we drop the empty poll function and the scheduler entry for it from the plugin? In my opinion its not needed at all.

https://github.com/smarthomeNG/plugins/blob/3ee29c118c920d0d413c8387d25710bf9cb7fa14/influxdb2/__init__.py#L249-L271

PR #857, opened 2023-11-24
modified some plugins for parameter checks, restartability and code cleanup

Morg42:


Morg42:
As I can't test these modifications myself, please test if plugin(s) run and work, if you use these

@Tom-Bom-badil as one maintainer

PR #889, opened 2023-12-30
smartVISU plugin widgets: fix import paths and remove obsolete imports

wvhn:


wvhn:
sorry. conflicts after syncing forks.

PR #918, opened 2024-03-03
plugin byd_bat : V0.1.3

lgb-this:
bug fixing
web interface log data in 2 tabs
additional log data text added


msinn:
Why does your pull request contain over 100 commits, most by other users and the oldest commit from over 6 years ago?

lgb-this:
No idea - i tried to add a 2nd plug-in "v-zug", but this was not possible with the same fork. I assume, that with this trial i created this mistake.

PR #929, opened 2024-03-31
AVM: Bugfix get_calllist and get_contact_name_by_phone_number

laraveluser:
Unfortunately, after the last update, some AVM functions no longer work as before. As this appears to be a typing error, please accept the bugfix

### DIFF
--- a/avm/__init__.py
+++ b/avm/__init__.py
@@ -595,7 +595,10 @@ class AVM(SmartPlugin):
 
     @NoAttributeError
     def get_calllist(self, filter_incoming: str = ''):
-        return self.fritz_device.get_calllist_from_cache()
+        if filter_incoming :
+            return self.fritz_device.get_calllist(filter_incoming)
+        else
+            return self.fritz_device.get_calllist_from_cache()
 
     @NoAttributeError
     def get_phone_name(self, index: int = 1):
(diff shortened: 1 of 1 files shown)


avm/__init__.py
```diff
@@ -594,7 +594,7 @@ def set_call_origin(self, phone_name: str):
         return self.fritz_device.set_call_origin(phone_name)
 
     @NoAttributeError
-    def get_calllist(self):
+    def get_calllist(self, filter_incoming: str = ''):
```
psilo909, line 597: is filter_incoming used somewhere? from what I see here, the executed function is not using it?

psilo909 (commented):
the rest looks fine, just check the one remark please.

avm/__init__.py
laraveluser: I honestly don't know why this definition is necessary, but get_calllist is defined twice in init.py and in the second function this filter is used and in my logic I have this filter in use - it had worked so far. However, it is no longer listed in the current documentation, so this is probably not a bug, but a change. If not, the executed method may also need to be adapted.

https://github.com/smarthomeNG/plugins/blob/7e00e4ee1f4ceea915bc0fb0628c10b92502cc8d/avm/__init__.py#L1435

Translated with DeepL.com (free version)

laraveluser: I have adapted it again so that both the new cache method and the previous query work. See below.

PR #945, opened 2024-07-03
plugins: adjust all plugins using suspend_item to new pause_item

Morg42:


Morg42:
This PR needs merged https://github.com/smarthomeNG/smarthome/pull/657 as a prerequisite

PR #981, opened 2024-11-27
Add plugin for piusv

SergeoLacruz:
This is for the piusv+.
The original SW for piusv+ is not up to date any longer. A open source version in C is available here:

https://github.com/dehapama/piupsmonitor-raspberrypi

The plugin communicates directly with the PIUSV via I2C. There is still some work to do but the major
functions are OK. Fine tuning is still needed.


Morg42:
Anyone able to test this?

piusv/__init__.py
```diff
@@ -0,0 +1,241 @@
...
+from .webif import WebInterface
+
+import smbus
+
+class piusv(SmartPlugin):
+    """
+    Main class of the Plugin. Does all plugin specific stuff and provides the update functions for the items
+    """
+    # Handle
+    piusv_handle = smbus.SMBus(1)
```
Morg42, line 40: Does instantiation connect to SMBus? In that case please initialize as None and intantiate in run(). Otherwise, I'd also recommend moving this to init to prevent class-members to span multiple instances (cleaner code)

wvhn:
Do I need to install the new piupsmon version in order to run the plugin? I can test the plugin if it runs with the original versio n.

SergeoLacruz:
The plugin does not need any piupsmon. I communicates with piusv via I2C. You can disable piupsmon.

wvhn:
Thanks for the plugin! It has been on my to-do list for a long time but I'm not really familiar with Python (yet).

Configuration docs are missing but I got it running. Had to activate i2c in raspi-config and add the user "smarthome" to the "i2c" group.
Now I'm getting errors in the log every few seconds:
```2024-12-15  15:25:09 ERROR    cherrypy.error.1721689952 [15/Dec/2024:15:25:09] HTTP 
> Traceback (most recent call last):
>   File "/home/smarthome/.local/lib/python3.9/site-packages/cherrypy/_cprequest.py", line 659, in respond
>     self._do_respond(path_info)
>   File "/home/smarthome/.local/lib/python3.9/site-packages/cherrypy/_cprequest.py", line 718, in _do_respond
>     response.body = self.handler()
>   File "/home/smarthome/.local/lib/python3.9/site-packages/cherrypy/lib/encoding.py", line 223, in __call__
>     self.body = self.oldhandler(*args, **kwargs)
>   File "/home/smarthome/.local/lib/python3.9/site-packages/cherrypy/_cpdispatch.py", line 54, in __call__
>     return self.callable(*self.args, **self.kwargs)
>   File "/usr/local/smarthome/plugins/piusv/webif/__init__.py", line 109, in get_data_html
>     data['plugin_suspended'] = self.plugin.suspended
> AttributeError: 'piusv' object has no attribute 'suspended'

The Webif is working well, however. It would be nice to see the voltage / current values in their physical units and an explanation of the status code would be great.

**Comment by SergeoLacruz** [comment]
I added the missing status variable. Surprisingly I did not get this error. 
The docu needs some work. You can create a visu panel like this:

grafik

The status value is the raw value from the piusv. You are right, it should be translated:
Bit 0 Micro-USB Power active
Bit 1 Uext Power active
Bit 2 Battery low
Bit 3 Battery charging
Bit 4 ?
Bit 5 Button 1 pressed
Bit 6 ?
Bit 7 ?

**Comment by wvhn** [comment]
Works now without error. Thanks!
Is there a reason why battery voltage is v_batt and not u_batt like u_ext and u_rasp?

**Comment by wvhn** [comment]
Starts now without error. Thanks!
- items are accessible and provide plausible data. Poll interval seems to be quite long. No information available how poll interval can be configured
- this should be moved to plugin config, too:
``` # setup scheduler for device poll loop   (disable the following line, if you don't need to poll the device. Rember to comment the self_cycle statement in __init__ as well) ```
- automatic updates in web interface are not working. You alway habe to aupdate manually. Update interval changes back to 30 sec if you try to change it 
- Is there a reason why battery voltage is v_batt and not u_batt like u_ext and u_rasp?

**Comment by SergeoLacruz** [comment]
I added the webinterface update. The function was missing. v_batt it renamed to u_batt. The poll cycle can be configured using the poll_cycle parameter in the plugin config. This should be in the automatically generated docs. Anyhow I need to work on the documentation. Disableing  the poll makes no sense. I will remove the comment. 
There is still some work to do for the next release :-)

**Comment by wvhn** [comment]
Super. Function is OK now. 
Using u_batt instead of v_batt is not really important but may help avoiding surprises in visu configuration. Thanks for that.

Maybe it's worth noting: I tested on a Raspi 2 with 32bit Bullseye system and shNG 1.9.5.

From my point of view, the PR can be merged into develop.

**Comment by SergeoLacruz** [comment]
Thanks for testing. I use a Pi4 with 64bit system.

**Comment by wvhn** [comment]
grafik

What about this widget for visualization? I could upload it as plugin-specific widget.

**Comment by SergeoLacruz** [comment]
Looks nice. Thanks for the contribution.

**Comment by wvhn** [comment]
I had some trouble with the table on small devices. This is now solved.
There is a PR into your repository which provides the widget in the required folder structure.

**Comment by SergeoLacruz** [comment]
Is here still something to do?

**Comment by wvhn** [comment]
I have been getting several errors during the day while I2C baud rate was still system default :
```
2025-01-10  01:08:26 ERROR    plugins.piusv       get_parameter: error writing to piusv
2025-01-10  02:56:38 ERROR    plugins.piusv       get_parameter: error reading to piusv
2025-01-10  03:33:02 ERROR    plugins.piusv       get_parameter: error reading to piusv
2025-01-10  08:38:23 ERROR    plugins.piusv       get_parameter: error writing to piusv
2025-01-10  08:58:37 ERROR    plugins.piusv       get_firmware: error reading to piusv
2025-01-10  08:58:37 WARNING  lib.item.item       Item usv.piusv_firmware: value "0" does not match type str. Via piusv None
2025-01-10  11:15:06 ERROR    plugins.piusv       get_status: error writing to piusv
2025-01-10  13:03:18 ERROR    plugins.piusv       get_parameter: error reading to piusv
```
Tests with baud rate 40000 are running now. 
From my perspective, only documentation is missing but you mentioned there is still some work to do ;)

SergeoLacruz:
I am also running 40000. No errors here.
I would like to close this now.

wvhn:
Error are gone after setting i2c baud rate to 40000.

@Morg42 : Plugin can be merged into develop

PR #985, opened 2024-12-13
Openweathermap: improve beaufort handling

onkelandy:


onkelandy:
Needs testing with correct API access

PR #1011, opened 2025-09-16
KNX Plugin: update type check due to xmltodict no longer returning OrderedDict

CaeruleusAqua:
fix knx plugin: update type check due to xmltodict no longer returning OrderedDict

Previously, we checked if `middle_dict` was an instance of `OrderedDict`:
```
if isinstance(middle_dict, OrderedDict):
```
However, per [xmltodict issue #252](https://github.com/martinblech/xmltodict/issues/252), recent versions of xmltodict no longer return an OrderedDict but standard dicts.

To accommodate this change, update the condition to:
```
if not isinstance(middle_dict, list):
```
which correctly handles cases where a `x_dict` is a dict object and to a list. ;-)

This ensures compatibility with the updated xmltodict behavior and prevents potential bugs caused by the old type check.

### DIFF
--- a/knx/knxproj.py
+++ b/knx/knxproj.py
@@ -265,7 +265,7 @@ def _parse_knxproject(filename, password=None):
         top = ga_dict.get('GroupRange',None)
         if top is None: return GAs
 
-        if isinstance(top, OrderedDict):
+        if not isinstance(top, list):
             # if there is only one child defined in xml, an ordered dict is
             # returned here instead of a list
             # so we need to convert it to a list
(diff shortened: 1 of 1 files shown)


bmxp (changes requested):
@CaeruleusAqua thank you for the PR and the findings!

There are three plugins that use this ``xmltodict``:

- plugin 'knx'
- plugin 'hue2' (@msinn ?)
- plugin 'sonos'  (@sisamiwe, @aschwith ?)
defining ``xmltodict>=0.11.0`` in their requirements.

We should update requirements for all three plugins and ensure that they still work when prescribing 1.0.x as requirement.
And we should review the code comments that ordered dict is not included any more.
Since support of knxproj is still at an early stage there should be test cases too for different ETS versions like 5 or 6 with also empty *default* knxproj files or ones including strange configurations.

PR #1013, opened 2025-09-18
Update requirements.txt of modbus_tcp

CaeruleusAqua:
The plugin currently requires a pymodbus version smaller than 3.10

### DIFF
--- a/modbus_tcp/requirements.txt
+++ b/modbus_tcp/requirements.txt
@@ -1,1 +1,1 @@
-pymodbus>=3.5.2;python_version>='3.8'
+pymodbus<3.10;python_version>='3.8'


bmxp:
Again pymodbus introduces breaking changes. It looks as if 3.10 introduced bugs and 3.11 reverts them.

Since a couple of plugins use pymodbus module we should inspect all requirements (again) and see if we can handle all plugin requirements:

```
# plugin 'trovis557x'
pymodbus>=2.2.0

# plugin 'helios_tcp'
pymodbus>=2.3,<3.0;python_version<'3.8'

# plugin 'helios_tcp'
pymodbus>=3.0.2;python_version>="3.8"

# plugin 'sma_mb'
# plugin 'pluggit'
# plugin 'priv_modbus_tcp2'
# plugin 'modbus_tcp'
# plugin 'ksemmodbus'
# plugin 'kostalmodbus'
pymodbus>=3.5.2;python_version>='3.8'
```

Morg42:
Changed all pymodbus requirements to '<3.10'.

Testing this without further plugin configuration, I get the following errors (cleaned for pure config errors):

python 3.10-3.13:
```
E 1016 094916 lib.plugin          | __init__                  ln:746  Plugin 'trovis' error importing Python package: No module named 'serial'  ----  Main
E 1016 094916 lib.plugin          | __init__                  ln:747  Plugin 'trovis' initialization failed, plugin not loaded  ----  Main
E 1016 094916 lib.plugin          | __init__                  ln:746  Plugin 'helios' error importing Python package: No module named 'pymodbus3'  ----  Main
E 1016 094916 lib.plugin          | __init__                  ln:747  Plugin 'helios' initialization failed, plugin not loaded  ----  Main
```
These seem to be missing in the respective requirements.

After installing pyserial for trovis, I get:

```
E 1016 100239 lib.plugin          | __init__                  ln:746  Plugin 'trovis' error importing Python package: cannot import name 'version' from 'pymodbus' (/Volumes/McDaten/sh/src/github/smarthome/venvs/py_3.13/lib/python3.13/site-packages/pymodbus/__init__.py)  ----  Main
E 1016 100239 lib.plugin          | __init__                  ln:747  Plugin 'trovis' initialization failed, plugin not loaded  ----  Main
```
This seems to be an error in trovis, which should be resolvable.


[edit] The plugin calling itself "helios" in the first log is actually "helios_tcp".

Tom-Bom-badil:
Some clarifications:
- **_Trovis_** can be ran in 2 operation modes: Modbus RTU (serial, traditional Modbus) or Modbus TCP. This is dependent on the adaptor you are using at the device. There is a config parameter which mode is used. Only the serial mode requires the serial library - I am not aware of anyone still running this mode, which also requires things like a `socat` daemon in addition.

- **_Helios_** should not throw any pymodbus related errors, as it is implementing its own protocol, also through _serial_. It is NOT needing pymodbus at all, so I wonder why it is mentioned here. There is another plugin `Helios TCP` which is meant for different (newer) devices with a different (newer) protocol - so maybe somthing got mixed up here. Original imports see [here](https://github.com/Tom-Bom-badil/smarthomeng_helios-vallox/blob/ed8343b064d85a7c2acbff39593a0d6c7a7fe40f/__init__.py#L22).

/tom

Morg42:
trovis should then possibly abandon serial support. If full functionality only can be achieved with pyserial, this should be in requirements. As requirements parsing cannot depend on plugin configuration, shng wouldn't start properly and possibly not even alert to the missing dependency.

You're right concerning helios, the actual plugin was helios_tcp, which isn't starting anyway in its current state.

Both plugins were only mentioned and might have to be acted on separately.

Thanks for your clarification!

Tom-Bom-badil:
For Trovis: I think it will be easier to just add "serial" to the requirements, instead of changing the code by removing serial.

I do not have an up-and-running shNG instance anymore, so I couldn't properly test such significant changes myself. Also, "not being aware of anyone still using the serial mode" doesn't necessarily mean that there aren't any users of it - there have been many in the early days of the plugin, and I cannot guarantee all users have switched over in the mean time.

/tom

PR #1039, opened 2026-06-03
database: add invalid-value handling and refactoring

Morg42:
As mentioned in #373, when a plugin (e.g.) loses the connection to a device, the item value becomes invalid, as it can not be ensured that is still the actual value. Items don't provide for saving this information, so an item saved to the database keeps its value.

This PR implements an item method `item.db_mark_invalid()`, which can be called by a plugin on (e.g.) connection loss. The database plugin registers this and starts a "value gap" until the item is updated with a new value or manually set valid again (`item.db_mark_valid()`). Plugins usually only need to call `item.db_mark_invalid()`.

Database aggregate functions are adjust so that invalid time periods are not counted and don't mess up averages or integral values.


Morg42:
for reference: commits were squash-merged directly into develop. Code and features are implemented.

PR #1054, opened 2026-09-11
smartvisu page generation: new working directory and stylesheet name

wvhn:
- moved the template files for page generation to ./dropins/shtemplates which is not in the loader path -> less conflicts
- renamed stylesheet from visu.css to shstyles.css in order to avoid conflicts


bmxp:
Do you have error check in place if copy/move actions fail due to access/rghts problems? Just in case one did not assign correct access rights..?

wvhn:
I used the same type of error checks that were already present in the plugin for comparable operations.

bmxp:
Ok, I do not use the generation. @Morg42 can you merge it?

wvhn:
Just wait, please. I tested again and found I have to correct one issue.

wvhn:
Done. Code can now be merged.

